PR #17, opened 2019-03-14
Implementation of modular design 

### LINKED ISSUE #15: Modular design
https://github.com/emersion/maddy/issues/15#issuecomment-471542210

--
_Original post:_

Let's say I configured IMAP endpoint as follows (where first line creates IMAP backend):
```
imap://127.0.0.1:1993 {
    sql sqlite3 maddy.db
}
```

Then I want to have SMTP endpoint that will deliver mail to same storage. How would I do this?

1. Assuming that backend also provides implementation of SMTP upstream (perhaps as separate object). 
```
smtp://127.0.0.1:1025 {
	sql sqlite3 maddy.db
}
```

This approach creates another set of problems, because it now requires two separate backend/upstream objects to coordinate access to the same storage (think of IMAP unilateral updates).
Global variables? External IPC sockets? All this seems to be dirty solution.

What is we can create one "storage" object and associate it with multiple IMAP/SMTP endpoints?
This will transform "another set of problems" into just serialization of access to storage object. Which is easily solved by throwing some mutexes into it (or even without them, I haven't tested that but go-sqlmail backend object should be safe for concurrent use by multiple goroutines).

It also reduces resources usage (we will have only one SQLite "connection" page cache, for example)

Now I can imagine something like this:
```
backend sql arbitrary_name {
  driver sqlite3
  dsn maddy.db
}

imap://127.0.0.1:1993 {
  backend arbitrary_name 
  # of course this requires storage object to implement go-imap's Backend interface
}

smtp://127.0.0.1:1025 {
  backend arbitrary_name 
  # and also go-smtp's Backend here now.
}
```

What do you think?

foxcpp:
As described in #15. Submitted here for sake of "early review".
Closes #15.


foxcpp:
Regarding separate repository.

maddy/module and maddy/config are moved into separate repository [maddy-lib](https://github.com/foxcpp/maddy-lib).
Otherwise, we get circular dependency:
- go-sqlmail needs maddy/module (for DeliveryContext struct)
- maddy needs go-sqlmail to use as a module
While this is not a problem without usage of Go modules, we use them and thus get circular dependency on module level.

There are two possible things to do:
- Merge go-sqlmail tree into maddy's and then continue development as a single project.
- Leave maddy-lib as is.

@emersion, what should I do?

foxcpp:
That's a lot of commits already, I guess. Here is my TODO list before I decide that this PR is ready:
In short, I want to reimplement most functionality of upstream maddy but using my proposed design.

- [x] Upgrade config parser to allow arbitrary nesting
- [x] Support env-var expansion in new config parser
- [x] Support snippets in new config parser
- [x] Support single imports in new config parser
- [ ] Support glob imports in new config parser
- [ ] Allow to omit instance module in configuration and assign something like dkim, dkim2, etc by default
- [x] Integrate go-sqlmail as a default backend (for now, this is not the best default, I know)
- [ ] Add tests for config parser
- [x] Implement configurable SMTP pipeline
- [ ] Add tests for SMTP pipeline code
- [x] Provide default SMTP pipeline
- [ ] Allow replacing delivery targets in SMTP pipeline without overriding the whole pipeline
- [ ] Reimplement DKIM verifier as a filter
- [ ] Add Received header to messages

Things I intentionally leave out of scope:
- Transparent PGP. This needs support for plugging arbitrary go-smtp backends into the SMTP pipeline, which is difficult to implement currently. Alternatively, it can be implemented as a filter. But I consider the whole idea questionable (as I said in #10) anyway.
- SMTP proxy. Not easily implementable with current SMTP pipeline design.
- IMAP proxy. Possible, but I don't consider this feature top-priority. Probably will be implemented in follow-up PRs.

emersion:
This seems like a pretty good plan. Do you think it's possible to break down the PR into smaller chunks? A +1.7k PR is already a little big to review.

>Reimplement DKIM verifier as a filter

I have plans to add functions to go-msgauth for this. Current design is: add a `msgauthgen` package to the library that returns a list of `Authentication-Results` that can be prepended to the header (DKIM and DMARC). Or do you have a better idea?

>SMTP proxy. Not easily implementable with current SMTP pipeline design.

Hmm. SMTP proxy is important for e.g. forwarding to LMTP sockets. Any reason why it can't be a `deliver` directive?

foxcpp:
> This seems like a pretty good plan. Do you think it's possible to break down the PR into smaller chunks? A +1.7k PR is already a little big to review.

I guess it is possible to split it like this:
* bf6d8eb5db683325ca1ddaa4fe3f8af604b9b5a9 - 4da37959c2c29cf79b1cce538b0fc1629baa05ad
  Basic components of module-based design. Only IMAP is implemented (but without backends, so it is  no-op actually).
* 83dd5cd09f92972cc99a5e9182b534bd380d3d9e
  New configuration parser.
* d07a46fca8a9ade9c6a7dd307111762e79ec5a9e
  Use go-sqlmail as a default IMAP backend.
* 3c57de16f00f4896356b03117ef098ef2cf3b2cd - 7b12fed8f78a4396d855f1cce51e727322d18f20
  SMTP pipeline implementation.
* e54887cb099c40190bd4765ff69ad02866e10e98 - 6706860411f4ccc9d78fdca88b3730df6d21a921
  Small tweaks for a lot of stuff.
* 4a9a756089634c4975109ed310837eb8b554ed36
  Documentation for everything above.

Also, some commits are somewhat out of order. I may try to sort them using rebase but I guess this will mess up a lot of things since almost all commits depend on previous ones. Considering all that I guess I should stop adding new code and just let you review these 1.7K lines.

> > Reimplement DKIM verifier as a filter
> 
> I have plans to add functions to go-msgauth for this. Current design is: add a `msgauthgen` package to the library that returns a list of `Authentication-Results` that can be prepended to the header (DKIM and DMARC). Or do you have a better idea?

Well, DKIM filter needs to the same thing (but only for DKIM (?)). So your `msgauthgen` package can help here.

> > SMTP proxy. Not easily implementable with current SMTP pipeline design.
> 
> Hmm. SMTP proxy is important for e.g. forwarding to LMTP sockets. Any reason why it can't be a `deliver` directive?

I wasn't sure about how to handle authentication. But now I can think we can just save credentials and pass them to the upstream server.

foxcpp:
Okay, I will not add more commits until you understand what I was doing in the previous 26.

README.md
```diff
@@ -7,32 +7,136 @@ Inspired from [Caddy](https://github.com/mholt/caddy).
 ## Features
 
 * Easy configuration with Maddyfile (same as Caddyfile)
-* Automatic TLS and DKIM support
 * Transparent PGP support
```
emersion: I guess we can remove this as well. Might just remove the whole "Features" section altogether.

README.md
```diff
@@ -7,32 +7,136 @@ Inspired from [Caddy](https://github.com/mholt/caddy).
...
+  Disable SQLite3 support in go-sqlmail (enabled by default). Saves around 9
+  MiB of binary size.
+* `mysql`
+  Include support for MySQL driver in go-sqlmail.
+* `postgresql`
+  Include support for PostgreSQL driver in go-sqlmail.
 
+## Configuration
+
+#### Syntax
```
emersion: This is using one heading level more than necessary

README.md
```diff
@@ -7,32 +7,136 @@ Inspired from [Caddy](https://github.com/mholt/caddy).
...
+    configuration_directives
 }
+```
+
+`instance_name` is the unique name of the configuration block. It is used when
+you need to refer to the module from different place in configuration (e.g.
+configure SMTP to deliver mail to certain specific storage)
+
+You can omit `instance_name`. If there is only one module config. block - it
+will get name the same as `module_name`. If there are multiple config blocks
```
emersion: Hmm, maybe just disallow having multiple modules without a name? It becomes very difficult to track down which module instance is which, especially with includes.

README.md
```diff
@@ -7,32 +7,136 @@ Inspired from [Caddy](https://github.com/mholt/caddy).
...
+
+smtp smtp://0.0.0.0 smtps://0.0.0.0:587 {
+    tls cert_file pkey_file
+    hostname emersion.fr
 }
 ```
+Don't forget to use actual values instead of placeholders.
+
+With this configuration, maddy will create an SQLite3 database for messages in
+current directory and use it to store all messages.
```
emersion: In the long term we'll probably want this to be somewhere in `/var` by default? This is good enough for now though.

README.md
```diff
@@ -7,32 +7,136 @@ Inspired from [Caddy](https://github.com/mholt/caddy).
...
+
+Currently, maddy doesn't implement any form of automatic TLS like Caddy. But
+since we don't want to have insecure defaults we require users to either
+manually configure TLS or disable it explicitly.
+
+Valid variants of TLS config directive:
+
+Disable TLS:
+```
+tls off
```
emersion: Maybe add a comment `# this is insecure` just to make sure people don't copy-pasta

config/parse.go
```diff
@@ -0,0 +1,197 @@
...
+import (
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/mholt/caddy/caddyfile"
+)
+
+type CfgTreeNode struct {
```
emersion: Since this is the `config` package, no need to repeat ourselves with `Cfg` here.

Also having both "tree" and "node" is a little weird. Maybe `Node` would be enough?

config/parse.go
```diff
@@ -0,0 +1,197 @@
...
+	Name      string
+	Args      []string
+	Childrens []CfgTreeNode
```
emersion: Typo: `Children`

module/module.go
```diff
@@ -0,0 +1,50 @@
...
+
+// Global WaitGroup instance is used to ensure graceful shutdown of server.
+// Whenever module starts goroutine (except when it is short-lived) - it should
+// increment WaitGroup counter by 1. Correspondingly, it should call Done when goroutine
+// finishes execution.
+//
+// If module requires special clean-up on shutdown - it should implement io.Closer interface
+// for this. Close() method will be called on server shutdown in this case. This method
+// should tell all module-created goroutines to stop.
+var WaitGroup sync.WaitGroup
```
emersion: Hmm. Instead of having this global state, can't we require `Close` to wait for goroutines to have finished?

module/registry.go
```diff
@@ -0,0 +1,59 @@
+package module
+
+import (
+	"sync"
+)
+
+var modules = make(map[string]NewModule)
+var modulesLck sync.RWMutex
```
emersion: Typo: "lock"?

module/module.go
```diff
@@ -0,0 +1,50 @@
...
+
+	// InstanceName method reports unique name of this module instance.
+	InstanceName() string
+
+	// Module version. Reported in logs.
+	Version() string
+}
+
+// NewModule is function that creates new instance of module with specified name.
+type NewModule func(name string, cfg config.CfgTreeNode) (Module, error)
```
emersion: These generally have the `Func` suffix.

smtp.go
```diff
@@ -2,209 +2,334 @@ package maddy
...
+}
 
-	if tokens, ok := tokens["auth"]; ok {
-		var err error
-		be, err = newSMTPAuth(caddyfile.NewDispenserTokens("", tokens), be)
-		if err != nil {
-			return nil, err
+func (u SMTPUser) Send(from string, to []string, r io.Reader) error {
+	ctx := module.DeliveryContext{}
+	ctx.From = from
```
emersion: Can't this be initialized inside `module.DeliveryContext{}` braces?

smtp.go
```diff
@@ -2,209 +2,334 @@ package maddy
...
+	ctx.To = to
+	ctx.Anonymous = u.anonymous
+	ctx.AuthUser = u.username
+	ctx.SrcHostname = u.state.Hostname
+	ctx.SrcAddr = u.state.RemoteAddr
+	ctx.OurHostname = u.endp.domain
+	ctx.Ctx = make(map[string]interface{})
+
+	currentMsg := r
+	var buf bytes.Buffer
```
emersion: Can you add a TODO to run filters in parallel?

Maybe this will require filters to return a `io.WriteCloser` instead, to ensure they get a `ctx` updated by the previous filter.

smtp.go
```diff
@@ -2,209 +2,334 @@ package maddy
...
 		}
-	}
 
-	if tokens, ok := tokens["pgp"]; ok {
-		var err error
-		be, err = newSMTPPGP(caddyfile.NewDispenserTokens("", tokens), be)
-		if err != nil {
-			return nil, err
+		if r != nil {
+			if _, err := io.Copy(&buf, r); err != nil {
```
emersion: We should reset the buffer right before, in case the filter doesn't consume the whole `io.Reader`.

What if the filter returns the same `io.Reader` it's received? We probably don't want to copy in this case.

smtp.go
```diff
@@ -2,209 +2,334 @@ package maddy
...
+			endp.domain = entry.Args[0]
+		case "maxidle":
+			if len(entry.Args) != 1 {
+				return nil, errors.New("maxidle: expected 1 argument")
+			}
+			maxIdle, err = strconv.Atoi(entry.Args[0])
+			if err != nil {
+				return nil, errors.New("maxidle: invalid integer value")
+			}
+		case "maxmsgbytes":
```
emersion: Could use "size" instead, so we can add units later on (`max_message_size 1MB`).

smtp.go
```diff
@@ -2,209 +2,334 @@ package maddy
...
+				log.Printf("smtp %s: TLS is disabled, this is insecure configuration and should be used only for testing!", instName)
+				insecureAuth = true
+			}
+		case "insecureauth":
+			log.Printf("smtp %s: authentication over unencrypted connections is allowed, this is insecure configuration and should be used only for testing!", instName)
+			insecureAuth = true
+		case "iodebug":
+			log.Printf("smtp %v: I/O debugging is on!\n", instName)
+			ioDebug = true
+		case "local-delivery":
```
emersion: We need to decide what conventions we use for these. Caddy uses underscores, so that would give things like `insecure_auth` and `local_delivery`.

README.md
foxcpp: Ok.

config/parse.go
foxcpp: It was initially placed in the maddy package. Forgot to rename after move. Will rename to `Node`.

README.md
foxcpp: Will remove Features section altogether.

README.md
foxcpp: Ok. Will fix.

README.md
foxcpp: Right, will change to allow only one instance with "implicit" name.

module/module.go
foxcpp: Ok.

module/registry.go
foxcpp: I do have a bad habit of dropping letters from commonly used words) Will change to "lock".

smtp.go
foxcpp: Ok.

smtp.go
foxcpp: The pipeline is strictly linear. I'm not sure how parallelism would look like here at all since every next step can depend on previous.

smtp.go
foxcpp: Ok.

smtp.go
foxcpp: Ok. Also, found out that I forgot to include this option in CONFIG_REFERENCE.md.

smtp.go
foxcpp: Will change to underscores.

smtp.go
emersion: The idea would be to make the signature be:

```
Process(ctx, out io.WriteCloser) (io.WriteCloser, error)
```

This returns immediately. The incoming message will be written to the returned `io.WriteCloser` by maddy. The filtered message will be written to `out` by the filter.

This allows to chain calls to `Process`. The `openpgp` package uses the same interface.

foxcpp: Alright. Will add TODO comment for it.

module/module.go
foxcpp: Right, this is better. Will remove global WG and instead add as a field to modules that require it (endpoints only now?).

smtp.go
emersion: (BTW this is more about memory consumption than latency -- if you have a 10MB e-mail with attachments you need to keep in memory and copy around it'll be pretty intensive)

config/parse.go
```diff
@@ -0,0 +1,197 @@
...
+	Snippet   bool
+}
+
+type parseContext struct {
+	caddyfile.Dispenser
+	parens   int
```
sapiens-sapide: typo in `parens`

foxcpp: https://www.dictionary.com/browse/parens
Is it really a typo? If my English is that bad - which word (or form) I should use then?

sapiens-sapide: sorry, I thought you meant « parents »

foxcpp:
2.6K LOC is already way too much, so I guess I will not try to implement everything I planned here.
Waiting for your comments!

foxcpp:
I guess you would really want to squash all these commits before merging.

README.md
```diff
@@ -4,36 +4,148 @@ Fast, cross-platform mail server.
...
+## Configuration
+
+### Syntax
+
+Maddy uses configuration format similar (but not the same!) to Caddy's
+Caddyfile.  You may want to read [Caddyfile page from Caddy docs](https://caddyserver.com/docs/caddyfile).
+
+Notable differences from Caddy's format:
+* Directives can't span multiple lines
+* Environmental variables are not expanded
```
emersion: I'm used to say "environment variables", is this a typo or is it just me?

README.md
```diff
@@ -4,36 +4,148 @@ Fast, cross-platform mail server.
...
+### go-sqlmail: Database location
+
+If you don't like SQLite3 or don't want to have it in the current directory,
+you can override the configuration of the default module.
+
+See [go-sqlmail repository](https://github.com/foxcpp/go-sqlmail) for
+information on RDBMS support.
+
+```
+sqlmail default {
```
emersion: Maybe we can just name this "sql" since it's alreayd obvious we're working with e-mails

README.md
foxcpp: Not sure how I got used to typing "environmental", but yes, this is a typo.

README.md
foxcpp: I named it this way because this is how "subproject" is named (go-sqlmail) but guess I will rename it to go-imap-sql, so yeah, it should be just `sql` now.

imap.go
```diff
@@ -1,102 +1,182 @@
...
+		insecureAuth bool
+		ioDebug      bool
+	)
+
+	cfg := config.Map{}
+	cfg.Custom("auth", false, false, defaultAuthProvider, authDirective, &endp.Auth)
+	cfg.Custom("storage", false, false, defaultStorage, storageDirective, &endp.Store)
+	cfg.Custom("tls", true, true, nil, tlsDirective, &endp.tlsConfig)
+	cfg.Bool("insecure_auth", false, &insecureAuth)
+	cfg.Bool("io_debug", false, &insecureAuth)
```
emersion: `&ioDebug`?

emersion: TBH I'm not sure we want to expose this as a configuration option

imap.go
```diff
@@ -1,102 +1,182 @@
...
 
-			target := addr.Host + ":" + addr.Port
-			if addr.IsTLS() {
-				return imapproxy.NewTLS(target, nil), nil
-			} else {
-				return imapproxy.New(target), nil
-			}
-		}
-	}
+func (endp *IMAPEndpoint) Version() string {
```
emersion: I'm wondering whether we really need those

imap.go
foxcpp: Oops!

imap.go
```diff
@@ -1,102 +1,182 @@
...
-	pgplocal "github.com/emersion/go-pgpmail/local"
-	"github.com/mholt/caddy/caddyfile"
+	"github.com/emersion/maddy/config"
+	"github.com/emersion/maddy/log"
+	"github.com/emersion/maddy/module"
+
+	appendlimit "github.com/emersion/go-imap-appendlimit"
+	compress "github.com/emersion/go-imap-compress"
+	move "github.com/emersion/go-imap-move"
+	"github.com/foxcpp/go-sqlmail/imap/children"
```
emersion: We could probably move this to maddy.

I've also been considering adding an `ext` package to go-imap.

imap.go
foxcpp: Well, this is meant for debugging purposes. Perhaps we may want to ask users to enable it while providing logs in bug reports so we can know what clients do exactly.

imap.go
emersion: Ah, circular dependencies…

imap.go
foxcpp: We have at least one "out-of-tree" module, so version information tied to the modules can be useful for bug reports. But I guess we can really drop it and just report go-imap-sql's version in debug log.

module/registry.go
```diff
@@ -0,0 +1,68 @@
...
+		panic("Register: module with specified name is already registered: " + name)
+	}
+
+	modules[name] = factory
+}
+
+// GetMod returns module from global registry.
+//
+// Nil is returned if no module with specified name is registered.
+func GetMod(name string) FuncNewModule {
```
emersion: This could just be named `Get` since we're already in the `module` package

smtp.go
```diff
@@ -2,209 +2,391 @@ package maddy
...
-	if err != nil {
-		return nil, err
+type SMTPUser struct {
+	endp      *SMTPEndpoint
+	anonymous bool
+	username  string
+	state     *smtp.ConnectionState
+}
+
+func (u SMTPUser) Send(from string, to []string, r io.Reader) error {
```
emersion: I'd prefer these to take a pointer. It's too easy to be surprised by `u` not being effectively mutated when doing `u.thing = stuff` in a method.

smtp.go
```diff
@@ -2,209 +2,391 @@ package maddy
...
-		be, err = newSMTPAuth(caddyfile.NewDispenserTokens("", tokens), be)
-		if err != nil {
-			return nil, err
-		}
+	// Check if TLS connection state struct is poplated.
+	// If it is - we are using TLS.
+	if u.state.TLS.Version != 0 {
+		ctx.SrcProto = "ESMTPS"
+	} else {
+		ctx.SrcProto = "ESMTP"
```
emersion: What about LMTP?

smtp.go
```diff
@@ -2,209 +2,391 @@ package maddy
...
+	}
+
+	return SMTPUser{
+		endp:     endp,
+		username: username,
+		state:    state,
+	}, nil
+}
+
+func (endp *SMTPEndpoint) AnonymousLogin(state *smtp.ConnectionState) (smtp.User, error) {
```
emersion: We could probably make this fail if auth is required

smtp.go
```diff
@@ -2,209 +2,391 @@ package maddy
...
+			//TODO: DKIM verify
+			deliveryStep{t: localDelivery, opts: localOpts},
+		)
 	}
+
+	return nil
+}
+
+func (endp *SMTPEndpoint) Login(state *smtp.ConnectionState, username, password string) (smtp.User, error) {
+	if !endp.Auth.CheckPlain(username, password) {
```
emersion: We could probably make this fail if it's a submission endpoint

smtp.go
```diff
@@ -2,209 +2,391 @@ package maddy
...
-			// TODO: don't print Received on a single line
-			received := "Received:"
-			// TODO: from
-			received += " by " + hostname
-			received += " with ESMTP"
-			// TODO: for
-			received += "; " + time.Now().Format(time.RFC1123Z) + "\r\n"
-			// TODO: add comments with TLS information
+func addReturnPath(ctx module.DeliveryContext, msg io.Reader) io.Reader {
+	hdr := "Return-Path: " + formatAddr(ctx.From) + "\r\n"
```
emersion: Do we sanitize this to make sure it's doesn't contain `\n`?

(The SMTP server might parse `\n` as not-a-newline since the RFC requires `\r\n`)

foxcpp: Totally forgot about this function. I should remove it anyway. And no, we don't sanitize, I assume that go-smtp doesn't return "invalid" values (here: ones with line separators). Will fix.

smtp.go
foxcpp: This is handled by prepending `require_auth` step to the pipeline in "submission mode".

smtppipeline.go
```diff
@@ -0,0 +1,289 @@
...
+				match = true
+				break
+			}
+		}
+	case "from":
+		match = step.matches(ctx.From)
+	case "from_domain":
+		parts := strings.Split(ctx.From, "@")
+		if len(parts) != 2 {
+			match = false
```
emersion: This doesn't bail out if `len(parts) != 2`

smtppipeline.go
```diff
@@ -0,0 +1,289 @@
...
+		}
+		if step.matches(parts[1]) {
```
emersion: This could be written `match = step.matches(parts[1])`

sqlmail.go
```diff
@@ -0,0 +1,168 @@
...
+	back, err := imapsqlmail.NewBackend(driver, dsn, opts)
+	if err != nil {
+		return fmt.Errorf("sqlmail: %s", err)
+	}
+	sqlm.Backend = back
+
+	return nil
+}
+
+func (sqlm *SQLMail) Extensions() []string {
```
emersion: Maybe we should rename this to `IMAPExtensions` -- we might have things like `SMTPExtensions` in the future.

foxcpp: Makes sense, will rename.

submission.go
```diff
@@ -0,0 +1,99 @@
...
+				Message: "Malformed Data header",
+			}
+		}
+	} else {
+		log.Debugf("adding missing Date header to message from %s (%s)", ctx.SrcHostname, ctx.SrcAddr)
+		parsed.Header.Set("Date", time.Now().Format("Mon, 2 Jan 2006 15:04:05 -0700"))
+	}
+
+	var buf bytes.Buffer
+	if err := parsed.WriteTo(&buf); err != nil {
```
emersion: Argh, this will break DKIM (working on fixing it). But in any case, we shouldn't re-format the message here. As a workaround we can copy it early at the beginning of the function and seek the buffer here

foxcpp: Your new interface in go-message allows preserving the initial structure of the message header. So this should not be a problem when it will implemented, right? I don't have DKIM now so this really breaks nothing. I would prefer to leave the code as is and instead use updated go-message.

imap.go
emersion: Okay, that's fair.

emersion: (Keep in mind it'll leak passwords and sensitive data -- we should be careful about this. Maybe add a log line warning about it?)

submission.go
emersion: Hmm. I'm not a fan of re-formatting and copying things around all over the place, but I guess this can be addressed separately.

README.md
foxcpp: Now I renamed go-sqlmail to go-imap-sql and module is named just 'sql'.

foxcpp:
There are several ideas you may want to get implemented too, but I'm intentionally leaving out here for now because I want to finally get something merged and let you review small PRs instead of this mess.

#####  Change default module name logic
  The problem with the current approach is that it requires configuration writers to be aware of the inner workings of code: They need to know the right default names for modules. I want to allow to omit then in the configuration where possible.
  Take for example PAM auth. module (from [another branch][pam-branch] in my fork), currently you need to use it like this:
  ```
  pam default_auth
  imap imap://0.0.0.0:993
  ```
  See this extra `default_auth`? If we omit it, then module will be named just "pam" and it will not make it default then. The solution is too allow module to specify default instance name for itself. So pam module will register itself as `default_auth` unless we explicitly specify name.
  ```
  pam
  imap imap://0.0.0.0:993
  ```

##### ...
I had some other ideas but forgot them :\

emersion:
>I'm intentionally leaving out here for now because I want to finally get something merged and let you review small PRs

+1

>The solution is to allow modules to specify the default instance name for itself. 

Maybe we should generate a name depending on the interfaces implemented by the module?

>Require STARTTLS on MTA port (25)

Related: MTA-STS lookup, which probably should be the default instead? I think we'll break too many servers otherwise (?).

tls.go
```diff
@@ -11,33 +11,34 @@ import (
...
+				return nil, err
 			}
+			return &cfg, nil
+		}
+	case 2:
+		cfg := tls.Config{}
+		if cert, err := tls.LoadX509KeyPair(node.Args[0], node.Args[1]); err != nil {
+			return nil, err
+		} else {
+			cfg.Certificates = append(cfg.Certificates, cert)
```
emersion: `cfg` is lost here, never returned

foxcpp: Oops. Never tested that branch since config refactoring.

emersion:
Thanks for this massive work!

PR #22, opened 2019-03-30
PAM-based and local-only PAM-less authentication

### LINKED ISSUE #20: Add PAM authentication support
This will allow to reuse a lot of components developed for PAM.

###### Notes
* CGo dependency: libpam
* libpam does have a callback-based interface, it needs to be used with extreme care to not introduce memory vulnerabilities.

###### Problems to solve
* maddy needs to access the shadow database (/etc/shadow) on configurations using the local database. Should we just require maddy to be running as root (or with CAP_DAC_READ_SEARCH) or somehow isolate code working with this file?

Some example code can be found here: https://stackoverflow.com/questions/10910193/how-to-authenticate-username-password-using-pam-w-o-root-privileges
Except probably we want to define our own PAM service instead of using "su".

### LINKED ISSUE #21: Add Pure Go local authentication support (Unix)
Generally, we should try to not introduce a lot of C dependencies (#20). We can implement basic support for local authentication on Unix systems using /etc/passwd and /etc/shadow databases.

foxcpp:
Implemented as an external binary helper to isolate privileged code. This PR also adds a generic extauth module to allow easy extension using 3rd-party authentication helpers.

The current recommended approach for installation is to grant helper binary additional permissions (setgid shadow, setuid root or CAP_DAC_READ_SEARCH) and restrict it to be runnable only by maddy user.

I decided to write a PAM helper in C because 100% of its code should call C library anyway. The code in Go would be much messier because of callbacks.

Closes #21.
Closes #20.

### DIFF
--- a/CONFIG_REFERENCE.md
+++ b/CONFIG_REFERENCE.md
@@ -244,3 +244,26 @@ No-op module. It doesn't need to be configured explicitly and can be referenced
 using `dummy` name. It can act as a filter, delivery target, and auth.
 provider.  In the latter case, it will accept any credentials, allowing any
 client to authenticate using any username and password (use with care!).
+
+### 'extauth' module
+
+Module for authentication using external helper binary.
+It looks for binary named maddy-auth-helper in $PATH and uses it for
+authentication.
+
+Protocol is very simple:
+Username and password are written to stdin, adding `\n` to the end. If binary
+exits with 0 status code - authentication is considered successful. If status
+code is 1 - authentication is failed. If status code is 2 - other unrelated
+error happened. Additional information should be written to stderr.
+
+```
+extauth default_auth
+```
+
+Valid configuration directives:
+* `helper`
+  Location of the helper binary.
+
+* `debug`
+  Verbose log only for this module.
--- a/CONFIG_REFERENCE.md
+++ b/CONFIG_REFERENCE.md
@@ -267,3 +267,10 @@ Valid configuration directives:
 
 * `debug`
   Verbose log only for this module.
+
+### 'pam' module
+
+Same as 'extauth' module but looks for [maddy-pam-helper] binary by default.
+
+[maddy-pam-helper][cmd/maddy-pam-helper]
+
--- a/CONFIG_REFERENCE.md
+++ b/CONFIG_REFERENCE.md
@@ -270,7 +270,15 @@ Valid configuration directives:
 
 ### 'pam' module
 
-Same as 'extauth' module but looks for [maddy-pam-helper] binary by default.
+Same as 'extauth' module but looks for [maddy-pam-helper] binary by default
+which implements authentication using PAM stack.
+
+### 'shadow' module
+
+Same as 'extauth' module but looks for [maddy-shadow-helper] binary by default
+which implements authenticaiton using local shadow database (/etc/shadow).
 
 [maddy-pam-helper][cmd/maddy-pam-helper]
+[maddy-shadow-helper][cmd/maddy-shadow-helper]
+
 
--- a/cmd/maddy-pam-helper/README.md
+++ b/cmd/maddy-pam-helper/README.md
@@ -1,0 +1,70 @@
+## maddy-pam-helper
+
+External setuid binary for interaction with shadow passwords database or other
+privileged objects necessary to run PAM authentication.
+
+### Building
+
+It is really easy to build it using any GCC:
+```
+gcc main.c -lpam -o maddy-pam-helper
+```
+
+Yes, it is not a Go binary.
+
+
+### Installation
+
+maddy-pam-helper is kinda dangerous binary and should not be allowed to be
+executed by everybody but maddy's user. At the same moment it needs to have
+access to read-protected files. For this reason installation should be done
+very carefully to make sure to not introduce any security "holes".
+
+#### First method
+
+```shell
+chown maddy: /usr/bin/maddy-pam-helper
+chmod u+x,g-x,o-x /usr/bin/maddy-pam-helper
+```
+
+Also maddy-pam-helper needs access to /etc/shadow, one of the ways to provide
+it is to set file capability CAP_DAC_READ_SEARCH:
+```
+setcap cap_dac_read_search+ep /usr/bin/maddy-pam-helper
+```
+
+#### Second method
+
+Another, less restrictive is to make it setuid-root (assuming you have both maddy user and group):
+```
+chown root:maddy /usr/bin/maddy-pam-helper
+chmod u+xs,g+x,o-x /usr/bin/maddy-pam-helper
+```
+
+#### Third method
+
+The best way actually is to create `shadow` group and grant access to
+/etc/shadow to it and then make maddy-pam-helper setgid-shadow:
+```
+groupadd shadow
+chown :shadow /etc/shadow
+chmod g+r /etc/shadow
+chown maddy:shadow /usr/bin/maddy-pam-helper
+chmod u+x,g+xs /usr/bin/maddy-pam-helper
+```
+
+Pick what works best for you.
+
+### PAM service
+
+maddy-pam-helper uses custom service instead of pretending to be su or sudo.
+Because of this you should configure PAM to accept it.
+
+Here is minimal example using local passwd/shadow database for authentication:
+```
+#%PAM-1.0
+auth		required	pam_unix.so
+account		required	pam_unix.so
+session		required	pam_unix.so
+```
+Put it into /etc/pam.d/maddy.
--- a/cmd/maddy-pam-helper/README.md
+++ b/cmd/maddy-pam-helper/README.md
@@ -60,10 +60,6 @@ Pick what works best for you.
 maddy-pam-helper uses custom service instead of pretending to be su or sudo.
 Because of this you should configure PAM to accept it.
 
-Here is minimal example using local passwd/shadow database for authentication:
-```
-#%PAM-1.0
-auth		required	pam_unix.so
-account		required	pam_unix.so
-```
-Put it into /etc/pam.d/maddy.
+Minimal example using local passwd/shadow database for authentication can be
+found in [maddy.conf][maddy.conf] file.
+It should be put into /etc/pam.d/maddy.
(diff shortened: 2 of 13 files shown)


foxcpp:
builds.sr.ht definitely needs an ability to test commits after branch merge.

CONFIG_REFERENCE.md
```diff
@@ -244,3 +244,41 @@ No-op module. It doesn't need to be configured explicitly and can be referenced
 using `dummy` name. It can act as a filter, delivery target, and auth.
 provider.  In the latter case, it will accept any credentials, allowing any
 client to authenticate using any username and password (use with care!).
+
+### 'extauth' module
+
+Module for authentication using external helper binary.
+It looks for binary named maddy-auth-helper in $PATH and uses it for
```
emersion: TBH we should put this in `libexecdir`

cmd/maddy-pam-helper/main.go
```diff
@@ -0,0 +1,20 @@
+package main
+
+/*
+#cgo LDFLAGS: -lpam
+#cgo CFLAGS: -DCGO
```
emersion: Can we build with `-Wall -Wextra -Wpedantic -std=c11`?

cmd/maddy-pam-helper/main.c
```diff
@@ -0,0 +1,69 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <security/pam_appl.h>
+
+/*
+I really doublt it is a good idea to bring Go to the binary whose primary task
```
emersion: Typo "doubt"

Yeah, using a pure C file is a good idea here.

cmd/maddy-pam-helper/main.c
```diff
@@ -0,0 +1,69 @@
...
+is to call libpam using CGo anyway.
+*/
+
+struct pam_response *reply;
```
emersion: This can be `static`

cmd/maddy-pam-helper/main.c
```diff
@@ -0,0 +1,69 @@
...
+
+int conv_func(int num_msg, const struct pam_message **msg, struct pam_response **resp, void *appdata_ptr) {
```
emersion: This can be `static`

cmd/maddy-pam-helper/main.c
```diff
@@ -0,0 +1,69 @@
...
+    *resp = reply;
+    return PAM_SUCCESS;
+}
+
+int run() {
```
emersion: In C functions without arguments are written `int run(void)` (otherwise they accept any number of args).

cmd/maddy-pam-helper/main.c
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    ssize_t password_len = getline(&password, &password_buf_len, stdin);
+    if (password_len < 0) {
+        perror("getline password");
+        return 2;
+    }
+
+    // Cut trailing \n.
+    username[username_len-1] = 0;
```
emersion: This doesn't check `username_len > 0`

cmd/maddy-pam-helper/main.c
```diff
@@ -0,0 +1,69 @@
...
+
+    const struct pam_conv local_conv = { conv_func, NULL };
+    pam_handle_t *local_auth = NULL;
+    int status = pam_start("maddy", username, &local_conv, &local_auth);
+    if (status != PAM_SUCCESS) {
+        fprintf(stderr, "pam_start: %s\n", pam_strerror(local_auth, status));
+        return 2;
+    }
+
+    reply = malloc(sizeof(struct pam_response));
```
emersion: `malloc` can return NULL

cmd/maddy-pam-helper/main.go
foxcpp: It will emit a lot of warnings for cgo wrapper.

cmd/maddy-pam-helper/main.c
```diff
@@ -0,0 +1,69 @@
...
+            return 2;
+        }
+    }
+
+    status = pam_end(local_auth, status);
+    if (status != PAM_SUCCESS) {
+            fprintf(stderr, "pam_end: %s\n", pam_strerror(local_auth, status));
+            return 2;
+    }
+}
```
emersion: This is missing a `return` statement

cmd/maddy-pam-helper/main.go
foxcpp: And it turns out we need `-std=gnu18` just because I used getline.

emersion: >It will emit a lot of warnings for cgo wrapper.

Oh and add `-Werror` to the mix too

>And it turns out we need -std=gnu18 just because I used getline. Should I replace it?

No no, `getline` is POSIX. You just need to `#define _POSIX_C_SOURCE 200809L` at the top of the file before any `#include`.

(See the manpage)

emersion: >It will emit a lot of warnings for cgo wrapper.

It's not very sane to write a security-oriented program in C without these

emersion: We probably don't even need `-std=c11`, `-std=c99` will probably be enough

foxcpp: Will add `-Wno-error=unused-parameter` to make warnings from cgo wrappers non-fatal.

emersion: Oh, this one is very annoying. Yeah just add `-Wno-unused-parameter` so we don't even see warnings

cmd/maddy-pam-helper/README.md
```diff
@@ -0,0 +1,69 @@
...
+chown maddy:shadow /usr/bin/maddy-pam-helper
+chmod u+x,g+xs /usr/bin/maddy-pam-helper
+```
+
+Pick what works best for you.
+
+### PAM service
+
+maddy-pam-helper uses custom service instead of pretending to be su or sudo.
+Because of this you should configure PAM to accept it.
```
emersion: We could put this in a file to make it easy for package maintainers to install it

shadow/verify_shadow.go
```diff
@@ -0,0 +1,58 @@
...
+	"time"
+
+	"github.com/GehirnInc/crypt"
+	_ "github.com/GehirnInc/crypt/apr1_crypt"
+	_ "github.com/GehirnInc/crypt/md5_crypt"
+	_ "github.com/GehirnInc/crypt/sha256_crypt"
+	_ "github.com/GehirnInc/crypt/sha512_crypt"
+)
+
+const SecsInDay = 86400
```
emersion: No need to export this

foxcpp:
Wops.

cmd/maddy-pam-helper/main.c
```diff
@@ -0,0 +1,86 @@
...
+
+/*
+I really doubt it is a good idea to bring Go to the binary whose primary task
+is to call libpam using CGo anyway.
+*/
+
+static struct pam_response *reply;
+
+static int conv_func(int num_msg, const struct pam_message **msg, struct pam_response **resp, void *appdata_ptr) {
+    // Silence "unused parameter" warnings.
```
emersion: Can you add `-Wno-unused-parameter` to CFLAGS so that we can remove this and silence warnings?

foxcpp:
I guess I should get rid of that habit of commit rewriting and instead rearrange them right before merge. Sorry for the inconvenience.

cmd/maddy-pam-helper/main.c
```diff
@@ -0,0 +1,80 @@
...
+
+    const struct pam_conv local_conv = { conv_func, NULL };
+    pam_handle_t *local_auth = NULL;
+    int status = pam_start("maddy", username, &local_conv, &local_auth);
+    if (status != PAM_SUCCESS) {
+        fprintf(stderr, "pam_start: %s\n", pam_strerror(local_auth, status));
+        return 2;
+    }
+
+    reply = malloc(sizeof(struct pam_response));
```
emersion: Eh, why do we need to malloc here anyway? We can allocate the `struct pam_response` on the stack instead

shadow/read_shadow.go
```diff
@@ -0,0 +1,80 @@
...
+	if err := scnr.Err(); err != nil {
+		return res, err
+	}
+	return res, nil
+}
+
+func parseEntry(line string) (*Entry, error) {
+	parts := strings.Split(line, ":")
+	if len(parts) != 9 {
+		return nil, errors.New("read: malformed entry")
```
emersion: Can we use the package name instead in these errors? Otherwise just having "malformed entry" makes it hard to understand it's coming from the function reading `/etc/shadow`

shadow/read_shadow.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	res := &Entry{
+		Name: parts[0],
+		Pass: parts[1],
+	}
+	var err error
```
emersion: Nit: this can be declared right before it's needed

shadow/read_shadow.go
foxcpp: Somewhere deep in PAM code free() is called for response (what contradicts man page btw)
I get this when I try to use variable on stack:
```
==20147== Invalid free() / delete / delete[] / realloc()
==20147==    at 0x48389AB: free (vg_replace_malloc.c:530)
==20147==    by 0x4871D63: pam_vprompt (in /usr/lib/libpam.so.0.84.2)
==20147==    by 0x4871F3A: pam_prompt (in /usr/lib/libpam.so.0.84.2)
==20147==    by 0x486E367: ??? (in /usr/lib/libpam.so.0.84.2)
==20147==    by 0x5247810: pam_sm_authenticate (in /usr/lib/security/pam_unix.so)
==20147==    by 0x486CB5C: ??? (in /usr/lib/libpam.so.0.84.2)
==20147==    by 0x486C3AF: pam_authenticate (in /usr/lib/libpam.so.0.84.2)
==20147==    by 0x10934D: run (main.c:57)
==20147==    by 0x109416: main (main.c:78)
==20147==  Address 0x10c0b0 is 0 bytes inside data symbol "reply"
==20147== 
```

emersion: Oh, right. PAM is terrible.

shadow/read_shadow.go
foxcpp: Fixed.

emersion:
Thanks!

PR #24, opened 2019-04-01
Make IMAPEndpoint implement backend.BackendUpdater

### LINKED ISSUE #23: Duplicate incoming messages
Messages are delivered multiple times to local mailboxes.

Investigation revealed that sometimes maddy fails to acknowledge message delivery and instead just doesn't reponds with anything despite successfull delivery. Remote MTA timeouts and retries delivery later resulting in duplicates.

Server restart fixes this problem, but only for some time. It happens only for some messages (needs to be additionally checked) and always reproduced when corresponding SMTP session is replayed as-is.

### Configuration 
maddy with experimental queue implementation (https://github.com/foxcpp/maddy/commit/824fbdc0ed8042916950d382151bcaf75df1de80). Based on this commit from this repository: https://github.com/emersion/maddy/commit/2355cd958adf9347bf5f2f485436159e8b8dc58f
Using go-imap-sql with PostgreSQL driver.

CC @NamedKitten

foxcpp:
This requires underlying storage to implement it too.

_Not sure if it qualifies as a "trivial patch" so I decided to make a PR._

Closes #23.

### DIFF
--- a/imap.go
+++ b/imap.go
@@ -28,6 +28,7 @@ type IMAPEndpoint struct {
 	Auth      module.AuthProvider
 	Store     module.Storage
 
+	updater     imapbackend.BackendUpdater
 	tlsConfig   *tls.Config
 	listenersWg sync.WaitGroup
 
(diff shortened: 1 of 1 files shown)


emersion:
Thanks!

PR #28, opened 2019-04-02
Improve configuration parser

### LINKED ISSUE #25: Improve configuration parser
Currently it is basically line-oriented and thus will not handle the following cases as expected:

* Closing brace will be ignored, leading to incorrect parsing of following statements.
  ```
  name args { directive }
  ```
  ```
  name args {
  directive }
  ```
  
* This will fail with parse error.
  ```
  name args 
  {
    directive
  }
  ```

foxcpp:
Closes #25.

### DIFF
--- a/config/parse.go
+++ b/config/parse.go
@@ -38,18 +38,18 @@ type parseContext struct {
 	fileLocation string
 }
 
-// readCfgNode reads node starting at current token pointed by lexer's
+// readNode reads node starting at current token pointed by lexer's
 // cursor (it should point to node name).
 //
-// After readCfgNode returns lexer's cursor will point to last token of parsed
+// After readNode returns lexer's cursor will point to last token of parsed
 // Node. This ensures predicatable cursor location independently of EOF state.
-// Thus code reading multiple nodes should call readCfgNode then manually
-// advance lexer cursor (ctx.Next) and either call readCfgNode again or stop
+// Thus code reading multiple nodes should call readNode then manually
+// advance lexer cursor (ctx.Next) and either call readNode again or stop
 // because cursor hit EOF.
 //
-// readCfgNode calls readCfgNodes if currently parsed node is a block.
+// readNode calls readNodes if currently parsed node is a block.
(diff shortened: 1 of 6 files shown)


config/parse.go
```diff
@@ -134,80 +96,125 @@ func (ctx *parseContext) isSnippet(name string) (bool, string) {
...
+
+	ctx.nesting++
+
 	for ctx.Next() {
+		// name arg0 arg1 {
+		//   c0
+		//   c1
+		// }
+		// ^ called when we hit } on separate line,
+		// This means block we are supposed to parse ended and we should stop.
```
emersion: "we are supposed to parse ended"

Typo here

config/parse.go
```diff
@@ -134,80 +96,125 @@ func (ctx *parseContext) isSnippet(name string) (bool, string) {
...
 	}
-	if ctx.parens < 0 {
-		return root.Children, ctx.snippets, ctx.Err("Unexpected }")
-	}
-	if ctx.parens > 0 {
-		return root.Children, ctx.snippets, ctx.Err("Unexpected {")
+
+	// There is no need to check ctx.nesting < 0 because it is checked by readNodes.
+	if ctx.nesting > 0 {
+		return root.Children, ctx.snippets, ctx.Err("unexpected {")
```
emersion: Isn't this "expected }" instead?

foxcpp: `ctx.nesting` is incremented every time we parse `{` and decremented every time we parse `}`.
So if after parsing `ctx.nesting` is negative - we have more }'s than {'s, if `ctx.nesting` is positive - we have more {'s than }'s. 
Though this error should be corrected to be "unexpected EOF" because it means we hit EOF when we were looking for }, /u didn't notice that when moving code around.

foxcpp: Changed error description (64ef51c).

config/parse.go
foxcpp: Fixed (64ef51c).

emersion:
Thanks!

PR #31, opened 2019-04-05
Properly integrate config.Map into code

foxcpp:
Another clean-up for #17. It makes several things slightly more clear.

PR #35, opened 2019-04-06
Change default file locations

### LINKED ISSUE #27: Replace default files location
By default configuration should be read from `/etc/maddy/Maddyfile`.
All state (queues, databases) should be stored in `$MADDYPATH` with the default value of `/var/lib/maddy`.

The current implementation is just a placeholder using the current working directory for everything.

foxcpp:
Now we use FHS directories by default. Depends on #31.

Closes #27.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,7 @@ are options that can be used like that:
   Default TLS certificate to use. See
   [CONFIG_REFERENCE.md](CONFIG_REFERENCE.md) for details.
 
+<<<<<<< HEAD
 * `debug`
   Write verbose logs describing what exactly is happening and how its going.
   Default mode is relatively quiet and still produces useful logs so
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -70,6 +72,22 @@ are options that can be used like that:
...
 
+* `statedir`
+  Change directory used for all state-related files.
+  Default is $MADDYSTATE environment variable or `/var/lib/maddy` if $MADDYSTATE is not set.
+  Default value can be changed using -X linker flag:
+  ```
+  go build --ldflags '-X github.com/emersion/maddy.defaultStateDirectory=/opt/maddy/state'
+  ```
+
+* `bindir`
```
emersion: This should be `libexecdir`

README.md
```diff
@@ -70,6 +72,22 @@ are options that can be used like that:
...
+  Change directory where all auxilary binaries are stored.
+  Default is $MADDYBIN environment variable or `/usr/libexec/maddy` if $MADDYBIN is not set.
```
emersion: Ditto

directories.go
```diff
@@ -0,0 +1,39 @@
...
+	defaultStateDirectory   = "/var/lib/maddy"
+	defaultLibexecDirectory = "/usr/libexec/maddy"
+)
+
+func ConfigDirectory() string {
+	return defaultConfigDirectory
+}
+
+func StateDirectory(globals map[string]interface{}) string {
+	if os.Getenv("MADDYSTATE") != "" {
```
emersion: Ah, one more thing: this is racy. Not a big deal, but something like this would be cleaner:

```go
if dir := os.Getenv("MADDYSTATE"); dir != "" {
	return dir
}
```

foxcpp: Is it possible to change the environment of the running process? [Googls says no](https://stackoverflow.com/a/8589266).

emersion: https://golang.org/pkg/os/#Setenv

foxcpp: Do we have any reason to change it from maddy? I don't so. Anyway, will change it because I don't like duplicated envvar name.

emersion:
Thanks!

PR #37, opened 2019-04-06
Lazy modules initialization

foxcpp:
Lazily initialize modules as they are requested. This fixes the initialization ordering issue introduced in fd14e52e459c31bda8974ec498bac77de283725a. Depends on #31.

### DIFF
--- a/maddy.go
+++ b/maddy.go
@@ -44,7 +44,7 @@ func Start(cfg []config.Node) error {
 			return fmt.Errorf("%s:%d: unknown module: %s", block.File, block.Line, modName)
 		}
 
-		if mod := module.GetInstance(instName); mod != nil {
+		if mod := module.GetUninitedInstance(instName); mod != nil {
 			return fmt.Errorf("%s:%d: module instance named %s already exists", block.File, block.Line, instName)
 		}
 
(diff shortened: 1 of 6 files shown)


maddy.go
```diff
@@ -44,7 +44,7 @@ func Start(cfg []config.Node) error {
 			return fmt.Errorf("%s:%d: unknown module: %s", block.File, block.Line, modName)
 		}
 
-		if mod := module.GetInstance(instName); mod != nil {
+		if mod := module.GetUninitedInstance(instName); mod != nil {
```
emersion: Maybe remove this and add `HasInstance` instead? Would be a safer API

module/instances.go
```diff
@@ -0,0 +1,69 @@
...
+
+import (
+	"fmt"
+
+	"github.com/emersion/maddy/config"
+	"github.com/emersion/maddy/log"
+)
+
+var (
+	Instances = make(map[string]struct {
```
emersion: This can be unexported

module/instances.go
```diff
@@ -0,0 +1,69 @@
...
+		mod Module
+		cfg *config.Map
+	})
+
+	Initialized = make(map[string]bool)
```
emersion: I'd prefer this to be unexported and expose a getter instead (`IsInitialized`?)

foxcpp: And then a setter too because maddy.Start needs to set status for modules it initializes.

emersion: Err. Okay, then it's fine to export this one, I guess

emersion:
Thanks!

PR #43, opened 2019-04-10
Drop most of the implicit defaults in favor of explicit configuration

foxcpp:
After some discussion on IRC, we agreed to drop a lot of implicitly implied configuration values. This includes default storage modules and SMTP pipeline definitions. I decided to leave some configuration values like timeouts as is (with reasonable defaults if nothing is specified in configuration).

I added the example configuration file to repository root. This configuration is for complete IMAP+MTA+MSA configuration with all implicit defaults used previously.


maddy.conf
```diff
@@ -0,0 +1,44 @@
+# Location of TLS certificate and private key. Global directive is used for all
+# endpoints.
+tls cert_file_path pkey_file
+# or if you are just testing maddy - you can run it without TLS.
+# however you also need to remove smtps:// and imaps:// listening addresses from
+# endpoints below.
+#tls off
```
emersion: I'd prefer not to have commented out parts. If we want to demonstrate different configurations for other use-cases, I'd rather have separate example files.

emersion:
Thanks!

Issue #32, opened 2019-04-05
Generic conditions and filtering for SMTP pipeline


foxcpp:
### Generic conditions syntax
```
value operator value
```

Value can be either variable prefixed with `$` or 

Variable can be *all fields from DeliveryContext* and also any key from DeliveryContext.Ctx. If the variable doesn't exist - its value assumed to be an empty string. 

The operator is one of the following:
* `=` **Variable value must be exactly equal to the value in condition.**
* `=*` Similar to the `=` but is case-insensitive.
* `!=` **Inverted `=`.**
* `!=*` Inverted `=*`.
* `~`  **Value is a regexp that should match variable value.**
* `~*` Case-insensitive version of `~`.
* `!~` **Inverted `~`.**
* `!~*` Inverted `~*`

Inspired by the [nginx's if]. I highlighted operations currently implemented for the `match` block.

Variable value is converted into string before comparsion.

#### Examples

```
$rcpt_domain = "emersion.fr"
$src_ip ~ "^192\.168\..+"
```

### `match` => `if`

Rename `match` block into `if` and make it use conditions syntax described above.

#### Examples

```
if $rcpt_domain = "emersion.fr" {
  proxy smtp://0.0.0.0
}
```

[nginx's if]: https://nginx.org/en/docs/http/ngx_http_rewrite_module.html#if

emersion:
>Generic conditions syntax
>match => if

LGTM

>This filtering is necessary due to the fall-through nature of the pipeline (deliver statement does not stop processing).

Ugh, don't we have a `stop` directive too?

>As a consequence, in the following example, maddy MTA will connect to itself and deliver email twice.

Hmm, I don't get why? If it takes a branch, it won't take the other?

>deliver [filter conditions] target [options]

Maybe it'll be worth it to use an alternate form to filter, e.g.:

```
deliver {
    filter <filter1> <filter2>
    target <target>
}
```

>Combined examples

Seems like there are duplicate conditions/filters here

foxcpp:
>Hmm, I don't get why? If it takes a branch, it won't take the other?

Let's say we are emersion.fr and we got a message with two recipients: foxcpp@foxcpp.dev and contact@emersion.fr.
And we have this pipeline definition:
```
submission smtps://0.0.0.0 {
  hostname emersion.fr
  if rcpt_domain = "emersion.fr" {
    deliver local
  }
  if rcpt_domain != "emersion.fr" {
    deliver remote
  }
}
```

[...]

PR #44, opened 2019-04-11
Implement destination pipeline step

foxcpp:
As described in https://github.com/emersion/maddy/issues/32#issuecomment-481628214.

#### CONFIG_REFERENCE.md
* `destination <recipient...> { ... }`

  Execute subblock only if message does have at least one recipient matching
  any of the specified rules.

  "Rule" can be either domain name, full address or regular expression.

  **Note:** If rule matches recipient - it will be removed from delivery context,
    thus no steps after this destination block will see this recipient.

  This ensures that the following example will not deliver message to both 

  Example: Deliver to "local" all messages for mailboxes on example.org and all other - to "dummy".
  ```
  destination example.org { delivery local }
  delivery dummy
  ```

### DIFF
--- a/CONFIG_REFERENCE.md
+++ b/CONFIG_REFERENCE.md
@@ -153,6 +153,9 @@ You can add any number of steps you want using following directives:
   Executes all nested steps if the condition specified in the directive is true
   (field contains the specified substring).
 
+  **Note:** `match` is a low-level dispatching primitive, in most cases you should
+  use `destination` instead.
+
   If the substring is wrapped in forward slashes - it will be interpreted as a
   Perl-compatible regular expression that should match field contents.
 
@@ -174,29 +177,35 @@ You can add any number of steps you want using following directives:
 
     Example:
     ```
-    match rcpt "/@emersion.fr$/" {
+    match rcpt_domain emersion.fr {
         filter dkim verify
         deliver local
     }
     ```
-
 
 * `destination <recipient...> { ... }`
 
   Execute subblock only if message does have at least one recipient matching
   any of the specified rules.
 
-  "Rule" can be either domain name, full address or regular expression.
+  "Rule" can be either domain name, full address (should include `@`) or regular expression.
 
   **Note:** If rule matches recipient - it will be removed from delivery context,
-    thus no steps after this destination block will see this recipient.
-
+  thus no steps after this destination block will see this recipient.
   This ensures that the following example will not deliver message to both 
+
+  **Note 2:** Don't forget that order of pipeline steps matters.
+  ```
+  deliver sql
+  destination postmaster@example.org { deliver local }
+  ```
+  In this case messages for postmaster@example.org will be delivered to
+  **both** 'sql' and 'local' storage.
 
   Example: Deliver to "local" all messages for mailboxes on example.org and all other - to "dummy".
   ```
-  destination example.org { delivery local }
-  delivery dummy
+  destination example.org { deliver local }
+  deliver dummy
   ```
 
 
--- a/CONFIG_REFERENCE.md
+++ b/CONFIG_REFERENCE.md
@@ -175,38 +175,27 @@ You can add any number of steps you want using following directives:
   - `src_hostname`
     Hostname reported by the client in the EHLO/HELO command.
 
-    Example:
-    ```
-    match rcpt_domain emersion.fr {
-        filter dkim verify
-        deliver local
-    }
-    ```
-
 * `destination <recipient...> { ... }`
 
-  Execute subblock only if message does have at least one recipient matching
-  any of the specified rules.
+  For all recipients that match at least one rule - execute subblock and stop
+  processing, for all others - skip block and continue.
 
-  "Rule" can be either domain name, full address (should include `@`) or regular expression.
-
-  **Note:** If rule matches recipient - it will be removed from delivery context,
-  thus no steps after this destination block will see this recipient.
-  This ensures that the following example will not deliver message to both 
-
-  **Note 2:** Don't forget that order of pipeline steps matters.
-  ```
-  deliver sql
-  destination postmaster@example.org { deliver local }
-  ```
-  In this case messages for postmaster@example.org will be delivered to
-  **both** 'sql' and 'local' storage.
+  "Rule" can be either domain name, full address (should include `@`) or
+  regular expression that should match full address.
 
   Example: Deliver to "local" all messages for mailboxes on example.org and all other - to "dummy".
   ```
   destination example.org { deliver local }
   deliver dummy
   ```
+
+  **Note:** Don't forget that order of pipeline steps matters.
+  ```
+  deliver sql
+  destination postmaster@example.org { deliver local }
+  ```
+  In this case messages for postmaster@example.org will be delivered to
+  **both** 'sql' and 'local' storage.
 
 
 Complete SMTP block example using custom pipeline:
--- a/CONFIG_REFERENCE.md
+++ b/CONFIG_REFERENCE.md
@@ -172,8 +172,35 @@ You can add any number of steps you want using following directives:
   - `src_hostname`
     Hostname reported by the client in the EHLO/HELO command.
 
-  See below for example.
+    Example:
+    ```
+    match rcpt "/@emersion.fr$/" {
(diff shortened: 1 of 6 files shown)


emersion:
I've thought about #32 for a while, and I can't find a better solution.

I'm a little worried about people using rules like `if rcpt = me@example.org { delivery local }`. I wonder how we should make it clear it won't work.

emersion:
This needs a rebase

foxcpp:
Rebased.

CONFIG_REFERENCE.md
```diff
@@ -194,25 +194,50 @@ specify them explicitly!)
   - `src_hostname`
     Hostname reported by the client in the EHLO/HELO command.
 
-  See below for example.
+    Example:
+    ```
+    match rcpt "/@emersion.fr$/" {
```
emersion: Ugh. I'm confused.

foxcpp: Right, I should replace it with rcpt_domain, forgot to do that when moving examples around.

CONFIG_REFERENCE.md
```diff
@@ -194,25 +197,56 @@ specify them explicitly!)
   - `src_hostname`
     Hostname reported by the client in the EHLO/HELO command.
 
-  See below for example.
+    Example:
+    ```
+    match rcpt_domain emersion.fr {
+        filter dkim verify
+        deliver local
```
emersion: So this will deliver locally e-mails that are for both local and non-local recipients?

CONFIG_REFERENCE.md
```diff
@@ -194,25 +197,56 @@ specify them explicitly!)
...
+    }
+    ```
 
+* `destination <recipient...> { ... }`
```
emersion: Thinking about it a little bit more, maybe we should name it `fork` or `split`? Not sure

smtppipeline.go
```diff
@@ -274,6 +255,116 @@ func matchStepFromCfg(node config.Node) (SMTPPipelineStep, error) {
...
+	return strings.ToLower(parts[1]) == strings.ToLower(cond.value)
+}
+
+type destinationStep struct {
+	conds    []destinationCond
+	substeps []SMTPPipelineStep
+}
+
+func (step *destinationStep) Pass(ctx *module.DeliveryContext, msg io.Reader) (io.Reader, bool, error) {
+	ctxCpy := *ctx
```
emersion: We should probably deep-copy all fields here. This would prevent filters from inside `destination` to mutate the context for consumers outside (e.g. if a filter replaces a recipient).

CONFIG_REFERENCE.md
foxcpp: And it is most likely to fail because the `local` module has no idea what to do with non-local recipients. Can you help me pick a better usage example here? (or should I remove it altogether?)

smtppipeline.go
foxcpp: Oh, right. Totally forgot that there are reference-based fields other than `To`.

CONFIG_REFERENCE.md
foxcpp: Does `destination` documentation sound like it splits something? I'm trying to hide the internal complexity of implementation, so the user will need to be aware of it only in cases where tight control of message dispatching is required.

foxcpp:
I guess we should get #43 merged first, because this PR requires some changes to default configuration and documentation to be complete. But I don't want to change things that are removed anyway in another PR (#43).

foxcpp:
Rebased, updated configuration file added in #43, also added a small change (d13000165e41597a6f8ed1cf61837e04654e2ee7).

CONFIG_REFERENCE.md
emersion: Okay, `destination` is probably better. However I don't think the documentation is clear enough. IMHO it should say something among the lines of: for all matching recipients execute the block and stop processing, for all others continue processing.

We may also want to make it mandatory to have `deliver` as the last line of a `destination` block.

CONFIG_REFERENCE.md
```diff
@@ -172,24 +175,56 @@ You can add any number of steps you want using following directives:
...
+  Execute subblock only if message does have at least one recipient matching
+  any of the specified rules.
+
+  "Rule" can be either domain name, full address (should include `@`) or regular expression.
 
+  **Note:** If rule matches recipient - it will be removed from delivery context,
+  thus no steps after this destination block will see this recipient.
+  This ensures that the following example will not deliver message to both 
+
+  **Note 2:** Don't forget that order of pipeline steps matters.
```
emersion: This note should probably come after the _good_ example

module/filter.go
```diff
@@ -58,6 +58,39 @@ type DeliveryContext struct {
 	Opts map[string]string
 }
 
+// DeepCopy method creates a copy of the DeliveryContext structure, also
```
emersion: "method" is unnecessary

module/filter.go
```diff
@@ -58,6 +58,39 @@ type DeliveryContext struct {
...
+//
+// There are two exceptions, however:
+// - SrcAddr is not copied and copys field references original value.
+// - Slice/map/pointer values in cpy.Ctx are not copied.
+func (ctx *DeliveryContext) DeepCopy() *DeliveryContext {
+	cpy := *ctx
+	// There is no good way to copy net.Addr, but it should not be
+	// modified by anything anyway so we are safe.
+
+	cpy.Ctx = make(map[string]interface{})
```
emersion: We could initialize the map with the correct capacity

module/filter.go
foxcpp: ops.

module/filter.go
foxcpp: What a good day, I didn't know that it is possible to set the capacity for a map.

CONFIG_REFERENCE.md
foxcpp: Yeah, that is much more explicit. 
Now it says:
```
  For all recipients that match at least one rule - execute subblock and stop
  processing, for all others - skip block and continue.
```

emersion: LGTM!

CONFIG_REFERENCE.md
```diff
@@ -172,24 +175,53 @@ You can add any number of steps you want using following directives:
...
+  "Rule" can be either domain name, full address (should include `@`) or
+  regular expression that should match full address.
+
+  Example: Deliver to "local" all messages for mailboxes on example.org and all other - to "dummy".
+  ```
+  destination example.org { deliver local }
+  deliver dummy
+  ```
+
+  **Note 2:** Don't forget that order of pipeline steps matters.
```
emersion: This is actually the first note :P

CONFIG_REFERENCE.md
emersion: I was about to suggest a silly example like "if the date matches YYY-12-24, add a X-Merry-Christmas header field" but we don't even have the required directives to do so. Probably better not to have an example for now.

smtppipeline.go
```diff
@@ -274,6 +255,117 @@ func matchStepFromCfg(node config.Node) (SMTPPipelineStep, error) {
...
+func passThroughPipeline(steps []SMTPPipelineStep, ctx *module.DeliveryContext, msg io.Reader) (io.Reader, bool, error) {
+	currentMsg := msg
+	var buf bytes.Buffer
+	for _, substep := range steps {
+		r, cont, err := substep.Pass(ctx, currentMsg)
+		if !cont {
+			return nil, cont, err
+		}
+
+		if r != nil {
```
emersion: Should also handle `r == &buf`

CONFIG_REFERENCE.md
foxcpp: doh

CONFIG_REFERENCE.md
emersion: Oh, can you remove this example?

foxcpp: Removed.

emersion:
Thanks!

foxcpp:
Just noticed that passThroughPipeline doesn't rewind reader buffer. :\

PR #45, opened 2019-04-11
Verify return-path for incoming messages

### LINKED ISSUE #41: Verify return-path for incoming messages
Following conditions should be met for verification to pass:
* rDNS domain of source server's IP should be equal to the hostname value presented in EHLO/HELO command
* Hostname domain presented in EHLO/HELO command should resolve to source server's IP address
* Domain in MAIL FROM should have MX record pointing to the source server

These checks are built on the assumption that we have a DNSSEC-enabled resolver and the source server does have DNSSEC enabled.

NamedKitten:
This should implement all the pipeline steps from issue #41 correctly.

Closes #41.

### DIFF
--- a/smtppipeline.go
+++ b/smtppipeline.go
@@ -24,6 +24,150 @@ type SMTPPipelineStep interface {
 	Pass(ctx *module.DeliveryContext, msg io.Reader) (newBody io.Reader, continueProcessing bool, err error)
 }
 
+type checkSourceHostnameStep struct {
+	required bool
+}
+
+func checkSourceHostnameStepFromCfg(node config.Node) (SMTPPipelineStep, error) {
+	required := false
+	if len(node.Args) >= 0 {
+		if node.Args[0] == "required" {
+			required = true
+		}
+	}
+
+	return checkSourceHostnameStep{
+		required: required,
+	}, nil
+}
+
+func (s checkSourceHostnameStep) Pass(ctx *module.DeliveryContext, _ io.Reader) (io.Reader, bool, error) {
+	ipAddr, ok := ctx.SrcAddr.(*net.IPAddr)
+	if !ok {
+		return nil, true, nil
+	}
+
+	srcIPs, err := net.LookupIP(ctx.SrcHostname)
+	if err != nil {
+		ctx.Ctx["src_hostname_check"] = false
+		if s.required {
+			return nil, false, errors.New("source hostname does not resolve")
+		} else {
+			return nil, true, nil
+		}
+	}
+
+	for _, ip := range srcIPs {
+		if ipAddr.IP.Equal(ip) {
+			ctx.Ctx["src_hostname_check"] = true
+			return nil, true, nil
+		}
+	}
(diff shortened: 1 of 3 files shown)


go.sum
```diff
@@ -0,0 +1,53 @@
+github.com/GehirnInc/crypt v0.0.0-20190301055215-6c0105aabd46/go.mod h1:kC29dT1vFpj7py2OvG1khBdQpo3kInWP+6QipLbdngo=
```
foxcpp: Remove go.sum file.
Hint: You can add it to .git/info/exclude so you will not accidentally commit it again.

@emersion, or perhaps we should allow it now?

smtppipeline.go
```diff
@@ -19,6 +19,30 @@ type SMTPPipelineStep interface {
...
+func (verifySourceHostnameStep) Pass(ctx *module.DeliveryContext, _ io.Reader) (io.Reader, bool, error) {
+	ipAddr, ok := ctx.SrcAddr.(*net.IPAddr)
+	if !ok {
+		return nil, true, nil
+	}
+
+	srcIPs, err := net.LookupIP(ctx.SrcHostname)
+	if err != nil {
+		ctx.Ctx["src_hostname_check"] = false
+		return nil, false, errors.New("Source hostname does not resolve")
```
foxcpp: Nitpick: [Error messages should not be capitalized](https://github.com/golang/go/wiki/CodeReviewComments#error-strings)

smtppipeline.go
```diff
@@ -19,6 +19,30 @@ type SMTPPipelineStep interface {
...
+	}
+
+	for _, ip := range srcIPs {
+		if ipAddr.IP.Equal(ip) {
+			ctx.Ctx["src_hostname_check"] = true
+			return nil, true, nil
+		}
+	}
+	ctx.Ctx["src_hostname_check"] = false
+	return nil, false, errors.New("Source hostname does not resolve to source address")
```
foxcpp: Same here.

smtppipeline.go
```diff
@@ -281,6 +305,8 @@ func StepFromCfg(node config.Node) (SMTPPipelineStep, error) {
 		return stopStep{}, nil
 	case "require_auth":
 		return requireAuthStep{}, nil
+	case "require_source_hostname":
```
foxcpp: Add documentation for the pipeline step to CONFIG_REFERENCE.md.

go.sum
foxcpp: Oops, didn't notice that you removed it in the next commit.

emersion: I'm fine with allowing it. (In any case not allowing it *and* not putting it in `.gitignore` is silly)

emersion: Should not be done in this PR though, it's unrelated

foxcpp:
Oops. That was a bad idea. I will fix it!

foxcpp:
NamedKitten's laptop broke so he can't finish it. I took care of it. I wonder who should review it now... @emersion?

emersion:
Thanks!

PR #48, opened 2019-04-13
Remove duplicated pipeline processing code

foxcpp:
It also solves the problem with missing buffer rewind I [noticed](https://github.com/emersion/maddy/pull/44#issuecomment-482894975)


emersion:
Thanks!

PR #49, opened 2019-04-14
Rework documentation

foxcpp:
CONFIG_REFERENCE.md is now maddy.conf(5) man page.
README.md is cleaned up.
Also, add maddy(1) man page.

Close #47.

PR #51, opened 2019-04-14
Implement basic queue with on-disk serialization

foxcpp:
For simplicity reasons, it uses one retry counter for all recipients.
It is just generic queue implementation, not [remote delivery](https://github.com/foxcpp/maddy/tree/remote-queue).

I already rebased it 13 times so I guess we need to finally get it upstream.
The implementation itself is actually almost [month old](https://github.com/emersion/maddy/issues/11#issuecomment-476012034), but I reworked it a week ago. 

Will need to be rebased (14th time!!!), when #49 will be merged.

### DIFF
--- a/maddy.conf.5.scd
+++ b/maddy.conf.5.scd
@@ -558,7 +558,7 @@ Valid configuration directives:
 
 File system directory to use to store queued messages.
 
-Default is /state_directory/configuration_block_name.
+Default is $MADDYSTATE/<configuration_block_name>.
 
 ## workers <count>
 
--- a/maddy.conf.5.scd
+++ b/maddy.conf.5.scd
@@ -539,6 +539,45 @@ Refuse to accept messages larger than `value` bytes. The default is 32 MiB.
 
 Enable verbose logging only for this configuration block.
 
+# QUEUE MODULE
+
+Queue module buffers messages on disk and retries delivery multiple times to
+ensure reliable delivery.
+
+```
+queue {
+    workers 16
+    max_tries 8
+    target dummy
+}
+```
+
+Valid configuration directives:
+
+## location
+
+File system directory to use to store queued messages.
+
+Default is /state_directory/configuration_block_name.
+
+## workers <count>
+
+Start up to 'count' goroutines for message processing. Basically, this option
+limits amount of messages tried to be delivered concurrently.
+
+Default is 16.
+
+## max_tries <count>
+
+Attempt delivery up to 'count' times. Note that no more attempts will be done
+is permanent error occured during previous attempt.
+
+Default is 4.
+
+## debug
+
+Enable verbose logging only for this configuration block.
+
 # EXTAUTH MODULE
 
 Module for authentication using external helper binary. It looks for binary
--- a/queue.go
+++ b/queue.go
@@ -81,10 +81,14 @@ func (q *Queue) Init(cfg *config.Map) error {
 	cfg.Bool("debug", true, &q.Log.Debug)
 	cfg.Int("max_retries", false, false, 4, &q.maxRetries)
 	cfg.Int("workers", false, false, 16, &workers)
-	cfg.String("location", false, true, "", &q.location)
+	cfg.String("location", false, false, filepath.Join(StateDirectory(cfg.Globals), q.name), &q.location)
 	cfg.Custom("target", false, true, nil, deliverDirective, &q.Target)
 	if _, err := cfg.Process(); err != nil {
 		return err
+	}
+
+	if !filepath.IsAbs(q.location) {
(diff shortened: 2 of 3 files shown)


maddy.conf.5.scd
```diff
@@ -539,6 +539,45 @@ Refuse to accept messages larger than `value` bytes. The default is 32 MiB.
...
+}
+```
+
+Valid configuration directives:
+
+## location
+
+File system directory to use to store queued messages.
+
+Default is /state_directory/configuration_block_name.
```
emersion: `$MADDYSTATE/<configuration_block_name>` would be better IMHO

timewheel.go
```diff
@@ -0,0 +1,115 @@
...
+				tw.slotsLock.Unlock()
+				tw.dispatch <- closestSlot
+
+				// break inside of select exits select, not outer loop
+				goto breakinnerloop
+			case newTarget := <-tw.updateNotify:
+				// Avoid unnecessary restarts if new target is not going to affect our
+				// current wait time.
+				if closestSlot.Time.Sub(now) <= newTarget.Sub(now) {
+					continue
```
emersion: TBH I'd just replace these `continue` with `goto`, the fact that `break`/`continue` won't behave the same way is too confusing

queue.go
```diff
@@ -0,0 +1,423 @@
...
+	return nil
+}
+
+func (q *Queue) worker() {
+	for slot := range q.wheel.Dispatch() {
+		q.Log.Debugln("worker woke up for", slot.Value)
+
+		id := slot.Value.(uuid.UUID)
+		meta, body, err := q.openMessage(id)
+		if err != nil {
```
emersion: `openMessage` doesn't close `body` in case of error

queue.go
```diff
@@ -0,0 +1,423 @@
...
+			q.Log.Printf("failed to read message: %v", err)
+			continue
+		}
+
+		q.Log.Debugln("try count", meta.TriesCount)
+
+		shouldRetry, newRcpts := q.attemptDelivery(&meta, body)
+		if !shouldRetry {
+			q.removeFromDisk(id)
+			continue
```
emersion: `body` is not closed in this case. Maybe it would be worth it to extract this logic into a function and `defer body.Close()`

queue.go
```diff
@@ -0,0 +1,423 @@
...
+		}
+		meta.Ctx.To = newRcpts
+
+		if meta.TriesCount == q.maxTries || len(meta.Ctx.To) == 0 {
+			q.removeFromDisk(id)
+			continue
```
emersion: Ditto

timewheel.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	tw.slotsLock.Lock()
+	tw.slots.PushBack(TimeSlot{Time: target, Value: value})
+	tw.slotsLock.Unlock()
+
+	tw.updateNotify <- target
+}
+
+func (tw *TimeWheel) Stop() {
```
emersion: Maybe this should be named `Close`

queue.go
```diff
@@ -0,0 +1,423 @@
...
+		}
+
+		meta, err := q.readMessageMeta(id)
+		if err != nil {
+			q.Log.Printf("failed to read meta-data for %s, skipping: %v", bodyName, err)
+			continue
+		}
+
+		nextTryTime := meta.LastAttempt
+		nextTryTime = nextTryTime.Add(q.initialRetryTime + q.initialRetryTime*time.Duration(math.Pow(q.retryTimeScale, float64(meta.TriesCount-1))))
```
emersion: Hmm, the first time this will be `q.initialRetryTime + q.initialRetryTime` instead of `q.initialRetryTime` right?

foxcpp: Woops.

foxcpp: Yeah, it should be just q.initialRetryTime. ce1b726

timewheel.go
foxcpp: ce1b726

queue.go
foxcpp: ce1b726

queue.go
foxcpp: ce1b726

queue.go
foxcpp: ce1b726

maddy.conf.5.scd
foxcpp: ce1b726

timewheel.go
foxcpp: To me, it looks like the comment above `break` makes everything clear and I don't like to create named entities (labels) here just for this.

emersion:
Thanks!

PR #52, opened 2019-04-17
ci: build man pages

emersion:


foxcpp:
It would be amazing to have man pages rendered to HTML and published at emersion.github.io/maddy or somewhere else, wdyt?

emersion:
Yeah. [ronn](http://rtomayko.github.io/ronn/ronn.1.html) seemed nice but it's written in Ruby and doesn't eat raw man pages. There are other options (that can probably be styled), need to investigate.

emersion:
Debian uses mandoc.

Issue #11, opened 2019-01-11
Implement a message queue


emersion:
This allows for multiple workers and makes sure no message gets lost.

foxcpp:
https://github.com/foxcpp/maddy/tree/queue

sapiens-sapide:
Is it really necessary ? Maddy could still have multiple workers to handle connections. Each workers could wait for the delivery completion before replying `250`. If a worker fails, mail delivery will be uncompleted and client will retry later.  

Not having a filesystem message queue would not only remove complexity from code, it'd also improve privacy ! ;-)

emersion:
What happens if your server hits a power outage? If maddy panics?

emersion:
>Each workers could wait for the delivery completion before replying 250.

Hmm. The reason other systems haven't done this is that it can take quite a while.

foxcpp:
>Each workers could wait for the delivery completion before replying 250. 

I already have seen multiple servers that reject your messages unless you try to submit them multiple times with some delay (hi, postgrey!).

sapiens-sapide:
Are we talking ingress or egress delivery ?

foxcpp:
Egress. For ingress, we already report any errors immediately.

emersion:
I guess this has been implemented in #51?

foxcpp:
Yep.

PR #55, opened 2019-04-27
Inline modules instances implementation

### LINKED ISSUE #42: Inline definitions of modules
Allow writing module definitions inside other module definitions, omitting names.

This feature would allow writing less verbose configuration files with fewer entities (so you don't have to look around a lot to understand what is happening). The key point is to require "named entities" only when they are shared between server components (modules).

For example, the following configuration:
```
hostname example.org

sql {
  driver sqlite3
  dsn /var/lib/maddy/all.db
}

smtp smtp://0.0.0.0:25 {
  pipeline incoming_smtp
}
smtp smtp://0.0.0.0:587 {
  pipeline outgoing_smtp
}

# let's pretend that we have two modules for DKIM signing and verification, dkim_sign and dkim_verify
dkim_sign {
  public_key pubkey_file_path
  private_key privkey_file_path
}

pipeline incoming_smtp {
    filter dkim_verify
    filter milter /var/run/spamassasin.sock
    deliver sql 
}

pipeline outgoing_smtp {
    filter dkim_sign
    deliver incoming_smtp local_only  # let's pretend that pipeline implements DeliveryTarget
    deliver queue remote_only
}

queue {
  max_tries 1
  target remote
}
```

could be written like that:
```
hostname example.org

sql {
  driver sqlite3
  dsn /var/lib/maddy/msgs.db
}

smtp smtp://0.0.0.0:25 {
  pipeline {
    filter dkim_verify require # refering to singleton module 'dkim_verify'
    filter milter /var/run/spamassasin.sock # refering to singleton module 'milter'
    deliver sql # refering to 'sql' module instance
  }
}

smtp smtp://0.0.0.0:587 {
  auth pam # refering to singleton module 'pam'
  pipeline {
    filter dkim_sign { # inline-defined instance of module dkim_sign
      public_key pubkey_file_path
      private_key privkey_file_path
    }
    deliver queue { # inline-defined instance of module queue
      max_tries 1
      target remote # refering to singleton module 'remote' here
    }
    deliver smtp
  }
}
```

Generic syntax change is from
```
{
  operation module_instance_name module_options
}
```
to
```
{
  operation module_name {
    module_config_directives
  }
}
```

Here "inner" pipeline instance gets some artificial name like `parentname_N` so it can use it as a key to store persistent data.
**!!!** Reordering of configuration directives here may change "artificial names". Can we do something about it?

**!!!** We need a different way to specify per-call options
Perhaps 
```
{
  operation module_name {
    module_config_directives
    opts options_go_here
  }
}
```
Note: `a { dir } b c` is parsed as 
```
a {
  dir
}
b c
```

But I'm not very sure about the whole idea. Any thougnts?

foxcpp:
Per-call module options are removed from the pipeline code as they are not necessary now.

Queue module from #11 should not be allowed to be used inline without explicit instance name or storage location.

Closes #42.

### DIFF
--- a/config.go
+++ b/config.go
@@ -14,90 +14,142 @@ import (
 Config matchers for module interfaces.
 */
 
-func authProvider(modName string) (module.AuthProvider, error) {
-	modObj, err := module.GetInstance(modName)
-	if err != nil {
-		return nil, err
+// createModule is a helper function for config matchers that can create inline modules.
+func createModule(args []string) (module.Module, error) {
+	modName := args[0]
+	var instName string
+	if len(args) >= 2 {
+		instName = args[1]
+		if module.HasInstance(instName) {
+			return nil, fmt.Errorf("module instance named %s already exists", instName)
+		}
+	}
+
+	newMod := module.Get(modName)
+	if newMod == nil {
+		return nil, fmt.Errorf("unknown module: %s", modName)
+	}
+
+	log.Debugln("module create", modName, instName, "(inline)")
+
+	return newMod(modName, instName)
+}
+
+func initInlineModule(modObj module.Module, globals map[string]interface{}, node *config.Node) error {
+	// This is to ensure modules Init will see expected node layout if it breaks
+	// Map abstraction and works with map.Values.
+	//
+	// Expected: modName modArgs { ... }
+	// Actual: something modName modArgs { ... }
+	node.Name = node.Args[0]
+	node.Args = node.Args[1:]
+
+	log.Debugln("module init", modObj.Name(), modObj.InstanceName(), "(inline)")
+	return modObj.Init(config.NewMap(globals, node))
+}
+
+func deliverDirective(m *config.Map, node *config.Node) (interface{}, error) {
+	return deliverTarget(m.Globals, node)
+}
+
+func deliverTarget(globals map[string]interface{}, node *config.Node) (module.DeliveryTarget, error) {
+	// First argument to make it compatible with config.Map.
+	if len(node.Args) == 0 {
+		return nil, config.NodeErr(node, "expected at least 1 argument")
+	}
+
+	var modObj module.Module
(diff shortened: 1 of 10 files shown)


smtppipeline.go
```diff
@@ -16,6 +16,9 @@ import (
 	"github.com/emersion/maddy/module"
 )
 
+// TODO: Consider merging SMTPPipelineStep interface with module.Filter and
+// converting check_source_* into filters.
```
emersion: I think this has been done

config.go
```diff
@@ -14,90 +14,104 @@ import (
...
 	}
 
-	backend, ok := modObj.(module.Storage)
-	if !ok {
-		return nil, fmt.Errorf("module %s doesn't implements storage interface", modObj.Name())
+	var modObj module.Module
+	var err error
+	if node.Children != nil {
+		modObj, err = createModule(node.Args)
+		if err != nil {
```
emersion: We can check this once after the `if` since it's duplicated in both branches.

Maybe this should logic be extracted in a function, it's used several times.

config/env.go
```diff
@@ -7,6 +7,10 @@ import (
 )
 
 func expandEnvironment(nodes []Node) []Node {
+	if nodes == nil {
```
emersion: Is there a reason why a `len` check isn't used here?

foxcpp: Added a comment to explain.

smtppipeline.go
foxcpp: Yep, changed the comment.

config.go
foxcpp: Done.

emersion:
Needs rebase

foxcpp:
https://builds.sr.ht/~foxcpp/job/61317

emersion:
Thanks!

PR #59, opened 2019-05-01
Extend tls directive to allow multiple certs and other options

### LINKED ISSUE #58: Per-domain TLS certificates
Perhaps something like that:
```
tls perdomain {
  <domain> <cert_file> <key_file>
  _default cert_file key_file
}
```

Possible using GetCertificate or GetConfigForClient callbacks in tls.Config.

foxcpp:
Closes #58. Using NameToCertificate map in tls.Config.
"Old" syntax (`tls cert key`) is still usable.


foxcpp:
So I just missed one important thing in #58. Should I remove logic for multiple certs from this PR?

emersion:
>So I just missed one important thing in #58. Should I remove logic for multiple certs from this PR?

I'd prefer that

foxcpp:
Removed.

tls.go
```diff
@@ -9,39 +9,178 @@ import (
...
+		return &cfg, nil
 	default:
 		return nil, m.MatchErr("expected 1 or 2 arguments")
 	}
 
 	return nil, nil
 }
 
+var strVersionsMap = map[string]uint16{
+	"SSL3.0": tls.VersionSSL30,
```
emersion: -1, this is insecure

emersion:
It's probably a good idea to use the same directives as https://caddyserver.com/docs/tls

maddy.conf.5.scd
```diff
@@ -255,6 +258,69 @@ go build --ldflags '-X github.com/emersion/maddy.defaultLibexecDirectory=/opt/ma
...
+- ECDHE-ECDSA-WITH-AES128-CBC-SHA256
+- ECDHE-RSA-WITH-AES128-CBC-SHA256
+- ECDHE-RSA-WITH-AES128-GCM-SHA256
+- ECDHE-ECDSA-WITH-AES128-GCM-SHA256
+- ECDHE-RSA-WITH-AES256-GCM-SHA384
+- ECDHE-ECDSA-WITH-AES256-GCM-SHA384
+- ECDHE-RSA-WITH-CHACHA20-POLY1305
+- ECDHE-ECDSA-WITH-CHACHA20-POLY1305
+
+## curve <curve0> [curve0] ...
```
emersion: `[curve1]`

tls.go
```diff
@@ -21,27 +23,171 @@ func tlsDirective(m *config.Map, node *config.Node) (interface{}, error) {
...
+	"ECDHE-ECDSA-WITH-CHACHA20-POLY1305": 0xcca9,
+
+	// TLS 1.3 cipher suites.
+	"AES128-GCM-SHA256":        0x1301,
+	"AES256-GCM-SHA384":        0x1302,
+	"CHACHA20-POLY1305-SHA256": 0x1303,
+}
+
+var strCurvesMap = map[string]tls.CurveID{
+	"p256":   23,
```
emersion: Can we use constants instead? (e.g. `tls.CurveP256`)

tls.go
```diff
@@ -21,27 +23,171 @@ func tlsDirective(m *config.Map, node *config.Node) (interface{}, error) {
...
+	"p384":   24,
+	"p521":   25,
+	"X25519": 29,
+}
+
+func readTlsBlock(blockNode *config.Node) (*tls.Config, error) {
```
emersion: Nit: `readTLSBlock`, see https://github.com/golang/go/wiki/CodeReviewComments#initialisms

tls.go
```diff
@@ -21,27 +23,171 @@ func tlsDirective(m *config.Map, node *config.Node) (interface{}, error) {
...
+	"tls1.0": tls.VersionTLS10,
+	"tls1.1": tls.VersionTLS11,
+	"tls1.2": tls.VersionTLS12,
+	"tls1.3": tls.VersionTLS13,
+	"":       0, // use crypto/tls defaults if value is not specified
+}
+
+var strCiphersMap = map[string]uint16{
+	// TLS 1.0 - 1.2 cipher suites.
+	"RSA-WITH-RC4128-SHA":                0x0005,
```
emersion: There are also constants for these (`tls.TLS_RSA_WITH_RC4_128_SHA`)

emersion: (I prefer to use constants because it would be unfortunate to map a cipher to a weaker one by mistake)

emersion:
Thanks!

PR #61, opened 2019-05-01
Add random DeliveryID to delivery context

foxcpp:
The ID is printed to logs and added to the Received header, so it is very useful for troubleshooting.

### DIFF
--- a/smtppipeline.go
+++ b/smtppipeline.go
@@ -257,7 +257,7 @@ func (step deliverStep) Pass(ctx *module.DeliveryContext, msg io.Reader) (io.Rea
 		}
 	}
 	received += fmt.Sprintf("\r\n\tby %s (envelope-sender <%s>)", sanitizeString(ctx.OurHostname), sanitizeString(ctx.From))
-	received += fmt.Sprintf("\r\n\twith %s", ctx.SrcProto)
+	received += fmt.Sprintf("\r\n\twith %s id %s", ctx.SrcProto, ctx.DeliveryID)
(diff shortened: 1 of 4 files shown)


emersion:
>added to the Received header

Where is this done?

foxcpp:
Wops, I'm way too careless today.

smtppipeline.go
```diff
@@ -257,7 +257,7 @@ func (step deliverStep) Pass(ctx *module.DeliveryContext, msg io.Reader) (io.Rea
 		}
 	}
 	received += fmt.Sprintf("\r\n\tby %s (envelope-sender <%s>)", sanitizeString(ctx.OurHostname), sanitizeString(ctx.From))
-	received += fmt.Sprintf("\r\n\twith %s", ctx.SrcProto)
+	received += fmt.Sprintf("\r\n\twith %s id %s", ctx.SrcProto, ctx.DeliveryID)
```
emersion: [Per RFC 5321](https://tools.ietf.org/html/rfc5321#page-59) "id" should be uppercase

foxcpp:
```
   ID             = CFWS "ID" FWS ( Atom / msg-id )
                  ; msg-id is defined in RFC 5322 [4]
```
Egh, isn't `Atom` just arbitrary string? (or if you are referring to "id" word before it - in ABNF quoted strings are case-insensitive ).

emersion: >ABNF strings are case-insensitive

Oh, right -- nevermind then

emersion:
Thanks!

PR #63, opened 2019-05-03
Update to the latest go-imap-sql version

foxcpp:
There are some backward-incompatible changes on `dev` branch of go-imap-sql. This should be merged once they are released (as 0.2, probably).


foxcpp:
I guess we are good to go. DB upgrade needs some manual steps (documented in go-imap-sql changelog).

emersion:
Thanks!

PR #65, opened 2019-05-11
Fix IPv6 address binding

lupine:
Consider this `maddy.conf` fragment

```
smtp smtp://[::]:25 {
    filter check_source_hostname

    destination example.com {
        deliver sql
    }
}
```

Starting maddy against this results in an error message:

```
11.05.19 15:50:43 failed to bind on smtp://::: listen tcp: address :::25: too many colons in address
```

With this change, the bind succeeds.

It's not my favourite way to fix this - I'd rather store `::` and do `JoinHostPort` in the right places - but that turned into a wider intervention than I was happy with for my first PR!

I also introduce `github.com/stretchr/testify`, and just a couple of tests for this. Let me know if that's inappropriate.


foxcpp:
https://github.com/emersion/maddy/blob/8beb8f9ea1c52d96b71a0471bde0f1a339f4e9f3/address.go#L79
Wouldn't use of net.JoinHostPort here fix the problem?

lupine:
On second thought, I'm really not happy with this. Give me a bit.

lupine:
@foxcpp it would, yes! That's the route I decided to go down instead. That's in https://github.com/emersion/maddy/pull/66

PR #66, opened 2019-05-11
Fix binding to IPv6 addresses

lupine:
Attempt #2 at https://github.com/emersion/maddy/pull/65 , this time with a better approach.

Both `Address.String()` and `Address.Address()` are modified to handle IPv6 literals appropriately.

### DIFF
--- a/address_test.go
+++ b/address_test.go
@@ -1,0 +1,33 @@
+package maddy
+
+import (
+	"net"
+	"reflect"
+	"testing"
+)
+
+func TestStandardizeAddress(t *testing.T) {
+	for _, expected := range []Address{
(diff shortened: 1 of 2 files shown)


address_test.go
```diff
@@ -0,0 +1,35 @@
+package maddy
+
+import (
+	"net"
+	"reflect"
+	"testing"
+)
+
+var cases = []Address{
```
foxcpp: Nitpicking, but can you move that slice into TestStandardizeAddress so we will not have too generically named variable ("cases", which cases?) in the global scope?

lupine: Done!

foxcpp:
Thanks!

PR #68, opened 2019-05-12
Implement support for remote delivery

foxcpp:
Some improvements are possible but let's start with this.

### DIFF
--- a/remote.go
+++ b/remote.go
@@ -20,9 +20,9 @@ import (
 var ErrTLSRequired = errors.New("TLS is required for outgoing connections but target server doesn't supports STARTTLS")
 
 type RemoteDelivery struct {
-	name             string
-	hostname         string
-	allowUnencrypted bool
+	name       string
+	hostname   string
+	requireTLS bool
 
 	Log log.Logger
 }
@@ -37,7 +37,7 @@ func NewRemoteDelivery(_, instName string) (module.Module, error) {
 func (rd *RemoteDelivery) Init(cfg *config.Map) error {
 	cfg.String("hostname", true, true, "", &rd.hostname)
 	cfg.Bool("debug", true, &rd.Log.Debug)
-	cfg.Bool("allow_unencrypted", false, &rd.allowUnencrypted)
+	cfg.Bool("require_tls", false, &rd.requireTLS)
 	if _, err := cfg.Process(); err != nil {
 		return err
 	}
@@ -109,7 +109,7 @@ func (rd *RemoteDelivery) Deliver(ctx module.DeliveryContext, msg io.Reader) err
 		var temporaryConnErr bool
 		var lastErr error
 		for _, host := range hosts {
-			cl, err = connectToServer(rd.hostname, host, !rd.allowUnencrypted)
+			cl, err = connectToServer(rd.hostname, host, !rd.requireTLS)
 			if err == nil {
 				usedHost = host
 				break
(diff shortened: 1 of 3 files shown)


remote.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+func isTemporaryErr(err error) bool {
+	if protoErr, ok := err.(*textproto.Error); ok {
+		return (protoErr.Code / 100) == 4
+	}
+	if smtpErr, ok := err.(*smtp.SMTPError); ok {
+		return (smtpErr.Code / 100) == 4
+	}
+	if dnsErr, ok := err.(*net.DNSError); ok {
```
emersion: It's probably safer to type assert to [`net.Error`](https://golang.org/pkg/net/#Error) instead

remote.go
```diff
@@ -0,0 +1,265 @@
...
+		return dnsErr.Temporary()
+	}
+
+	if strings.HasPrefix(err.Error(), "x509") {
```
emersion: Ugh

remote.go
```diff
@@ -0,0 +1,265 @@
...
+	if err != nil {
+		rd.Log.Printf("DATA failed: %v (server = %s, delivery ID = %s)", err, usedServer, ctx.DeliveryID)
+		return toPartialError(isTemporaryErr(err), rcpts, err)
+	}
+	defer bodyWriter.Close()
+	if _, err := io.Copy(bodyWriter, body); err != nil {
+		rd.Log.Printf("body write failed: %v (server = %s, delivery ID = %s)", err, usedServer, ctx.DeliveryID)
+		return toPartialError(isTemporaryErr(err), rcpts, err)
+	}
+
```
emersion: Can we close the `bodyWriter` here and check the error? (It should be fine to double-close and not check the second close error in `defer`)

remote.go
```diff
@@ -0,0 +1,265 @@
...
+	groupedRcpts, err := groupByDomain(ctx.To)
+	if err != nil {
+		return err
+	}
+
+	partialErr := PartialError{
+		Errs: make(map[string]error),
+	}
+	// TODO: look into ways to parallelize this, the main trouble here is body
+	// probably create pipe for each server and copy body to each?
```
emersion: Could also call `bytes.NewReader` for each server. We should be careful not to completely stall everything if one server is slow.

remote.go
foxcpp: Errors with `x509` prefix indicate certificate verification failure. I'm not sure whether we should consider them to be temporary. Wdyt? Anyway, somehow I missed that there is a concrete type for it (`x509.CertificateInvalidError`).

remote.go
foxcpp: Should I reimplement it using `bytes.NewReader` or leave it as is for now?

remote.go
emersion: 1. My "ugh" was about checking the error message. This should really be last-resort if we really need it.
2. Because of certificates expiring and/or misconfiguration, it's probably a good idea to retry anyway.

remote.go
emersion: I'm fine with the current behaviour, I prefer adding features that are simple and work and then improve them in future PRs.

remote.go
foxcpp: Will do.

remote.go
foxcpp: Done.

remote.go
foxcpp: Done. Removed this check.

emersion:
Thanks!

PR #74, opened 2019-05-24
Support for per-domain authentication & storage and user@domain login names

### LINKED ISSUE #57: Considerations for handling user@domain in username
**Main Use-Case:** Use different databases for different domains.

**Proposed Solution:** Introduce module `perdomain` that calls different auth. providers depending on the source domain and implement auth. provider interface itself.

```
auth perdomain {
  domain example.org {
    auth virtual { file /etc/maddy/example.org-users }
  }
  domain example.com {
    dont_strip_domain # by default perdomain removes domain from username before passing it to the "nested" provider
    auth virtual { file /etc/maddy/example.com-users }
  }

  # Both make authentication fail by default.
  unknown { # domain not matched by any directives.
    ...
  }
  nodomain { # no domain present in username
    ...
  }
}
```

foxcpp:
Note about 4e804b1: It was done to allow to override globally enabled boolean directive to false on the block level. I think it might be useful with boolean flags that can be set on the global level.

Closes #57.

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -1,0 +1,35 @@
+package maddy
+
+import "strings"
+
+func checkDomainAuth(username string, perDomain bool, allowedDomains []string) (loginName string, allowed bool) {
+	var accountName, domain string
+	if perDomain {
+		parts := strings.Split(username, "@")
+		if len(parts) != 2 {
+			return "", false
+		}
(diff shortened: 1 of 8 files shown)


emersion:
Thanks!

PR #81, opened 2019-05-26
Implement MTA-STS policy enforcing

### LINKED ISSUE #40: Enforce MTA-STS policy
Look-up MTA-STS policy for remote MTAs before connecting to them, enforce TLS and cache policy.

foxcpp:
Closes #40.

### DIFF
--- a/mtasts/cache.go
+++ b/mtasts/cache.go
@@ -3,12 +3,15 @@ package mtasts
 import (
 	"encoding/json"
 	"errors"
+	"io/ioutil"
 	"net"
 	"net/http"
 	"os"
 	"path/filepath"
 	"strings"
 	"time"
+
+	"github.com/emersion/maddy/log"
 )
 
 func downloadPolicy(domain string) (*Policy, error) {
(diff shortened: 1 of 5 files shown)


emersion:
Maybe some of this could be added to https://github.com/emersion/go-mtasts?

foxcpp:
Ouch, I should have checked whether you have a library for it before writing my own. mtasts.go (with corresponding tests) can be merged into go-mtasts. Not sure about Cache.fetch, probably it can be split and part not involving cache moved to go-mtasts.

foxcpp:
Alright, I will contribute parsing functions into go-mtasts.

emersion:
TBH the library is so small, I was wondering whether it's worth it to have a separate lib for this.

>It seems odd that your library uses JSON since RFC 8461 doesn't use it.

Ah, I think it implements a draft. Maybe the final RFC has diverged.

foxcpp:
On second thought, yeah, probably you are right. It is only 159 lines of code (+238 for tests).

mtasts/cache.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	defer resp.Body.Close()
+
+	// Policies fetched via HTTPS are only valid if the HTTP response code is
+	// 200 (OK).  HTTP 3xx redirects MUST NOT be followed.
+	if resp.StatusCode != 200 {
+		return nil, errors.New("mtasts: HTTP " + resp.Status)
+	}
+
+	if !strings.HasPrefix(resp.Header.Get("Content-Type"), "text/plain") {
```
emersion: We can use `mime.ParseMediaType`, this would be more robust

mtasts/cache.go
```diff
@@ -0,0 +1,174 @@
...
+		FetchTime time.Time
+		Policy    *Policy
+	}{}
+	if err := json.NewDecoder(f).Decode(&data); err != nil {
+		return "", time.Time{}, nil, err
+	}
+	return data.ID, data.FetchTime, data.Policy, nil
+}
+
+func (c *Cache) RefreshCache() error {
```
emersion: I'm wondering whether refreshing the cache like this is a good idea. An alternative would be to refresh only when we need the MTA-STS policy (ie. we're sending an e-mail) and our cache is out-of-date.

foxcpp: https://tools.ietf.org/html/rfc8461#section-10.2
See the third paragraph.

>   Because this attack is also possible upon refresh of a cached policy,
   implementers SHOULD NOT wait until a cached policy has expired before
   checking for an update; if senders attempt to refresh the cache
   regularly (for example, by fetching the current live policy in a
   background task that runs daily or weekly, regardless of the state of
   the "_mta-sts" TXT record, and updating their cache's "max age"
   accordingly), an attacker would have to foil policy discovery
   consistently over the lifetime of a cached policy to prevent a
   successful refresh.

emersion: Oh, that makes a lot of sense.

foxcpp: Though I missed that part about TXT record, it might be important to actually ignore it while refreshing cache.

foxcpp:
Rebased.

mtasts/cache.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+	}
+
+	if !validCache || dnsId != cachedId {
+		policy, err := downloadPolicy(domain)
+		if err != nil {
+			return false, nil, err
+		}
+
+		if err := c.store(domain, dnsId, time.Now(), policy); err != nil {
```
emersion: We should log the error

remote.go
```diff
@@ -126,17 +155,38 @@ func toPartialError(temporary bool, rcpts []string, err error) *PartialError {
...
 func (rd *RemoteDelivery) deliverForServer(ctx *module.DeliveryContext, domain string, rcpts []string, body io.Reader) *PartialError {
 	addrs, err := lookupTargetServers(domain)
 	if err != nil {
 		return toPartialError(false, rcpts, err)
 	}
 
+	stsPolicy, err := rd.mtastsCache.Get(domain)
+	if err != nil && err != mtasts.ErrNoPolicy {
+		rd.Log.Printf("failed to fetch MTA-STS policy for %s: %v", domain, err)
+		// Problems with policy should be threated as temporary ones.
```
emersion: Nit: typo in treated

foxcpp: I should definitely get a spell checker into my code editor.

emersion:
Thanks!

PR #89, opened 2019-06-01
Enable IMAP ID extension

foxcpp:


emersion:
TBH I'd rather not support this extension…

foxcpp:
>TBH I'd rather not support this extension…

Why?

emersion:
This encourages clients to have per-server quirks. And if by chance clients don't do that, this doesn't have any value.

Do you have a use-case in mind?

foxcpp:
Internet-wide scans for collecting statistics of used server software?

emersion:
Meh, I don't care how many people use maddy.

foxcpp:
Alright then. No ID.

PR #107, opened 2019-08-04
New message dispatcher

### LINKED ISSUE #32: Generic conditions and filtering for SMTP pipeline

### LINKED ISSUE #54: Pre-DATA checks on envelope
We want to run certain checks (#41) on the envelope before the message body is sent to us to not waste traffic. The current pipeline implementation doesn't allow to run steps before the message body is received.

Also, SMTP doesn't allow us to "abort" transaction once we sent 354 intermediate reply to the DATA command.

I see two possible solutions for this problem:

### Define a special form of checks to be run before the body is sent

Defined somehow like that:
```
smtp ... {
  envelope_checks {
    verify_source_hostname
    verify_source_mx
    ...
  }
}
```

The checks are executed when RCPT/MAIL FROM command received and use the following interface:
```go
type EnvelopeCheck interface {
  CheckFrom(ctx *DeliveryContext, from string) error
  CheckRcpt(ctx *DeliveryContext, rcpt string) error
}
```
Corresponding method is called for each recipient and for source.

#### Advantages

* Ability to report failures per recipient as a status for RCPT TO command.

#### Disadvantages

* See the Advantages of the next method.

### Start pipeline before the body is received

Once the client issues the DATA command, start executing pipeline steps. If any of the steps need the message body - send 354 reply and pass body reader to the pipeline step (with some tricks it can be made fully transparent for step implementation).

#### Advantages

* Uniform interface for all checks
* Conditional checks using pipeline dispatching mechanisms (such as `destination` directive)

#### Disadvantages

* Requires changes to go-smtp (pre-body callback for DATA command)
* It is not possible to report errors for individual recipients or even MAIL FROM command.
  Quoting RFC 5321:
  > Using a "550 mailbox not found" (or
   equivalent) reply code after the data are accepted makes it difficult
   or impossible for the client to determine which recipients failed.

Question regarding standards conformance/interoperability: Are we allowed to return arbitrary failures in the DATA command before the client sends us the body?

### LINKED ISSUE #80: Special-case postmaster address
```
RCPT TO:<postmaster>
```
Should be accepted as valid.
There are several places around the code that assume user@domain in recipients addresses, need to check them and probably move some code around.

See https://tools.ietf.org/html/rfc5321#page-14 and https://tools.ietf.org/html/rfc5321#section-4.5.1.

### LINKED ISSUE #101: Pipeline parallelism
Most pipeline steps  do network I/O (DNS, mostly) so sequential execution is very slow. But serial execution of pipeline directives is actually its feature: It is very flexible but still simple model, a minimal domain-specific language, roughly speaking. I think we want to keep it. Instead, add a meta-directive "parallel" to indicate that some block should be parallelized (think of pre-delivery filters/checks). What do you think? Or, perhaps it is possible to redesign it from scratch with parallelism in mind, I'm not sure how to do it though.

foxcpp:
Seeking for high level feedback on proposed design before I rewrite remaining code to use it.

- [x] Message buffering primitives
- [x] Per-source message routing
- [x] Per-recipient message routing
- [x] Message checking
- [x] Dispatcher config parser
- [x] Update dummy module
- [x] Update remote module
- [x] Update DNS checks modules
- [x] Update smtp module
- [x] Update queue module
- [x] Update sql module
- [ ] Implement base code for "message modificators"
- [x] Update the user documentation to reflect changes

Closes #101.
Closes #54.
Closes #80.
Closes #32 (no longer relevant).


foxcpp:
@emersion, do you think I should split that mess somehow or you can't review it thoroughly so it's not worth it anyway? (repeating my question from IRC)

foxcpp:
Okay, I ~~will~~can move queue/remote/sql rewrites into a separate PRs. I just need to split a few commits that affect multiple modules. Do you want me to do that?

foxcpp:
I will complete the remaining subtasks as a separate PRs, just to keep stuff reviewable.

Upd: nope, since repo is now mine, this would be unnecessary.

foxcpp:
I'm going to recreate this PR because GitHub seems to act weirdly because this PR was created from the removed fork, which is now upstream.

PR #124, opened 2019-08-30
Replace msg id with logger wrapper

amitlevy21:
@foxcpp I'm not entirely pleased with this.
The `dispatcherDeliveryLogger` function I add is almost exact like the original `deliveryLogger` function in stateless_check.go. It feels like I can avoid code duplication here.

You may wonder why did I add [this](https://github.com/amitlevy21/maddy/blob/35c1cf5b2a7ea0b6ed81cabad10187739ab012db/dispatcher.go#L49). I'm trying to avoid getting [here](https://github.com/amitlevy21/maddy/blob/35c1cf5b2a7ea0b6ed81cabad10187739ab012db/dispatcher.go#L72) and invoking l.out on a nil.

I'd like to have some feedback, before I continue and make the same changes to the rest of the files needed.

I was also wondering, what tests can I add to this? Adding tests that check log specific content would easily break.

Thanks

### DIFF
--- a/dispatcher.go
+++ b/dispatcher.go
@@ -67,13 +67,14 @@ func splitAddress(addr string) (mailbox, domain string, err error) {
 }
 
 func (d *Dispatcher) Start(msgMeta *module.MsgMetadata, mailFrom string) (module.Delivery, error) {
+	dl := deliveryLogger(d.Log, msgMeta)
 	dd := dispatcherDelivery{
 		d:               d,
 		rcptChecksState: make(map[*rcptBlock]module.CheckState),
 		deliveries:      make(map[module.DeliveryTarget]module.Delivery),
 		msgMeta:         msgMeta,
 		cancelCtx:       context.Background(),
-		log:             &d.Log,
+		log:             &dl,
 		checkScore:      0,
 	}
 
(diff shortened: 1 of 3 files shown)


Hexawolf:
@amitlevy21 you fucking crazy mate, your commit messages are becoming more and more bloated

.gitignore
```diff
@@ -22,3 +22,4 @@ _testmain.go
 *.exe
 *.test
 *.prof
+.vscode/
```
foxcpp: I would prefer to use .gitignore only for project-specific files. If you use a tool that creates something in project directory - add it to `.git/info/exclude` instead of `.gitignore`.

dispatcher.go
```diff
@@ -66,14 +68,29 @@ func splitAddress(addr string) (mailbox, domain string, err error) {
 	}
 }
 
+func dispatcherDeliveryLogger(l log.Logger, msgMeta *module.MsgMetadata) *log.Logger {
```
foxcpp: Rebase the your commit branch on upstream master and use the fixed deliveryLogger function.

Here are Git magic to do so, make sure your working directory is clean (you have no uncommited changes):
```
git remote add upstream https://github.com/foxcpp/maddy.git
git fetch upstream
git rebase upstream/master
```

After updating your code you will have to use `--force` flag for `git push` once to rewrite commits you already pushed.

dispatcher.go
```diff
@@ -45,6 +46,7 @@ func NewDispatcher(globals map[string]interface{}, cfg []config.Node) (*Dispatch
 	parsedCfg, err := parseDispatcherRootCfg(globals, cfg)
 	return &Dispatcher{
 		dispatcherCfg: parsedCfg,
+		Log: log.DefaultLogger,
```
foxcpp: Here is the problem: this Logger is replaced by module that embed dispatcher functionality.
This means that it will be replaced with module with possibly nil Out again.

As I said in https://github.com/foxcpp/maddy/pull/124/commits/35c1cf5b2a7ea0b6ed81cabad10187739ab012db#r319509126, it is enough to have fixed deliveryLogger version that uses log.DefaultLogger.Out if Loggers Out is nil. log.DefaultLogger.Out is always non-nil.

amitlevy21:
@foxcpp I'll squash using rebase once I'm done with the rest.

remote.go
```diff
@@ -114,6 +114,7 @@ func (rt *RemoteTarget) Start(msgMeta *module.MsgMetadata, mailFrom string) (mod
 }
 
 func (rd *remoteDelivery) AddRcpt(to string) error {
+	dl := deliveryLogger(rd.rt.Log, rd.msgMeta)
```
foxcpp: Can you add the logger field to remoteDelivery? It would be cleaner IMO.

foxcpp (commented):
LGTM otherwise.

foxcpp (approved):
LGTM

foxcpp:
Thanks for your work!

amitlevy21:
@foxcpp Thank you for your patience :)

Issue #133, opened 2019-09-12
require_tls check


foxcpp:
Connection check that rejects messages unless TLS is used to deliver them.

foxcpp:
Closed by #166.

PR #166, opened 2019-11-13
Add a requiretls check for incoming SMTP

lupine:
Implements #133 

Currently WIP as it's untested.

### DIFF
--- a/docs/man/maddy-filters.5.scd
+++ b/docs/man/maddy-filters.5.scd
@@ -35,7 +35,7 @@ storage, the storage backend can place the message in the 'Junk' mailbox.
 Another thing to keep in mind that 'remote' module (see *maddy-targets*(5))
 will refuse to send quarantined messages.
 
-# DNS checks
+# Simple checks
 
 ## Configuration directives
 
@@ -77,6 +77,14 @@ specified in EHLO/HELO command.
 
 By default, quarantines messages coming from servers with mismatched or missing
 PTR record, use 'fail_action' directive to change that.
+
+## require_tls
+
+Check that the source server is connected via TLS; either directly, or by using
+the STARTTLS command.
+
+By default, rejects messages coming from unencrypted servers. Use the
+'fail_action' directive to change that.
 
 # DKIM authentication module (verify_dkim)
 
--- a/internal/check/requiretls/requiretls.go
+++ b/internal/check/requiretls/requiretls.go
@@ -1,0 +1,26 @@
+package requiretls
+
+import (
+	"github.com/foxcpp/maddy/internal/check"
+	"github.com/foxcpp/maddy/internal/exterrors"
+	"github.com/foxcpp/maddy/internal/module"
+)
+
+func requireTLS(ctx check.StatelessCheckContext) module.CheckResult {
+	if ctx.MsgMeta.Conn != nil && ctx.MsgMeta.Conn.TLS.HandshakeComplete {
+		return module.CheckResult{}
+	}
+
+	return module.CheckResult{
+		Reason: &exterrors.SMTPError{
+			Code:         550,
+			EnhancedCode: exterrors.EnhancedCode{5, 7, 1},
+			Message:      "TLS conversation required",
+			CheckName:    "require_tls",
+		},
+	}
+}
+
+func init() {
+	check.RegisterStatelessCheck("require_tls", check.FailAction{Reject: true}, requireTLS, nil, nil, nil)
+}
--- a/maddy.go
+++ b/maddy.go
@@ -25,6 +25,7 @@ import (
 	_ "github.com/foxcpp/maddy/internal/check/dkim"
 	_ "github.com/foxcpp/maddy/internal/check/dns"
 	_ "github.com/foxcpp/maddy/internal/check/dnsbl"
+	_ "github.com/foxcpp/maddy/internal/check/requiretls"
 	_ "github.com/foxcpp/maddy/internal/check/spf"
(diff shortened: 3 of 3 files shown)


check/requiretls/requiretls.go
```diff
@@ -0,0 +1,26 @@
...
+
+func requireTLS(ctx check.StatelessCheckContext) module.CheckResult {
+	if ctx.MsgMeta.SrcTLSState.HandshakeComplete {
+		return module.CheckResult{}
+	}
+
+	return module.CheckResult{
+		Reason: &exterrors.SMTPError{
+			Code:         550,
+			EnhancedCode: exterrors.EnhancedCode{5, 7, 1},
```
foxcpp: More specific code for it is 5.7.10 and 523.

(from https://www.iana.org/assignments/smtp-enhanced-status-codes/smtp-enhanced-status-codes.xhtml)

lupine: @foxcpp I did see that status code, but I think it's inappropriate to use here. The explanatory text is:

> This indicates that external strong privacy layer is needed in order to use the requested authentication mechanism. This is primarily intended for use with clear text authentication mechanisms. A client which receives this may activate a security layer such as TLS prior to authenticating, or attempt to use a stronger mechanism.

I read that to mean that it's only expected as a response to the `AUTH` command.

I picked 550 (and guessed 5.7.1), and the error text, based on this thread about what exchange does: https://social.technet.microsoft.com/Forums/ie/en-US/3d6cb80a-687e-407d-8cb4-e5c313f92757/email-bouncing-back-8220denied-by-policy-tls-conversation-required-diagnostic-code-smtp-550?forum=exchangesvrsecuremessaging

Happy to change it if my guess about 5.7.10 is wrong, though!

foxcpp: Probably you are right. It is not a big deal, anyway.

maddy.conf
```diff
@@ -46,6 +46,9 @@ smtp tcp://0.0.0.0:25 {
         # Verify that domain in MAIL FROM does have a MX record.
         require_mx_record
 
+        # Verify that TLS or STARTTLS is being used by the client. Fail if it is not.
+        require_tls
```
foxcpp: I am slightly worried about that being in the default configuration, though. I think some legitimate messages are still delivered over plain-text connections, I do [not want to break it][1]. To mitigate downgrade attacks however (if sender server lacks the DANE/DNSSEC or MTA-STS support or it is not configured by the recipient), maddy might remember the use of TLS by the sender server and enforce it second time (like chasquid [does][2]).

[1]: https://github.com/foxcpp/maddy/blob/master/HACKING.md#design-goals
[2]: https://blitiri.com.ar/p/chasquid/docs/sec-levels/#incoming-connections

lupine: Yeah, it probably doesn't want to be on by default. I was wondering where best to add it to the manpage.

foxcpp: Perhaps add a "Connection state checks" section to maddy-filter(5) and then mention require_tls in a subsection?

check/requiretls/requiretls.go
```diff
@@ -0,0 +1,26 @@
...
+			Message:      "TLS conversation required",
+			CheckName:    "require_tls",
+		},
+	}
+}
+
+func init() {
+	check.RegisterStatelessCheck("require_tls", check.FailAction{Reject: true}, requireTLS, nil, nil, nil)
```
lupine: I have a doubt about making this a per-connection check.

Say a connection is made, the client tries to send a message, and it fails because `require_tls` is enabled. Can the client then run `STARTTLS` and attempt to re-send? Will the per-connection check be re-run at that point, or will the original failure continue to apply? Should this, instead, be a per-message check, even though it's only checking connection params?

check/requiretls/requiretls.go
```diff
@@ -0,0 +1,26 @@
+package requiretls
+
+import (
+	"github.com/foxcpp/maddy/check"
+	"github.com/foxcpp/maddy/exterrors"
+	"github.com/foxcpp/maddy/module"
+)
+
+func requireTLS(ctx check.StatelessCheckContext) module.CheckResult {
+	if ctx.MsgMeta.SrcTLSState.HandshakeComplete {
```
lupine: I expected `SrcTLSState` to be a pointer, so it could have a `nil` value. Checking the `HandshakeComplete` boolean works too, but I wonder if it might be better to refactor so it can be a pointer instead. WDYT?

foxcpp: No objection. Though, you might as well refactor it in the go-smtp, since maddy simply copies the field from ConnectionState struct passed to it.

check/requiretls/requiretls.go
foxcpp: STARTTLS resets the go-smtp session and the check will rerun (essentially, from the maddy perspective it will look like a new connection).

foxcpp: Though, CheckConnection is more of a logical rather than a technical separation. It still runs on the each message received over the same connection (though this might change in the future as an optimization).

foxcpp:
Needs rebase, see a28d5bcf44599a34f150ae5378ba280284ff186e.

maddy.conf
lupine: I've renamed "DNS checks", since it's fundamentally the same kind of thing. WDYT?

lupine: (I figure we could add tls preference memory in a later PR)

foxcpp (approved):
LGTM

PR #183, opened 2019-12-20
changed yes/no to accept different true/false type values.

letto4135:
Added different values to the acceptable yes/no per issue 181. Also handled the case of the input by changing to lowercase

### DIFF
--- a/internal/config/map.go
+++ b/internal/config/map.go
@@ -8,7 +8,6 @@ import (
 	"strings"
 	"time"
 	"unicode"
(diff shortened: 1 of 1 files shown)


internal/config/map.go
```diff
@@ -8,7 +8,8 @@ import (
 	"strings"
 	"time"
 	"unicode"
-
+	"strings"
```
foxcpp: It is already imported above.

PR #184, opened 2019-12-21
get.sh: fix go toolchain version mismatch

kk-boop:
Fixes go toolchain download version and unpack version mismatch, now both rely on the variable.


foxcpp:
Woops, thanks for the fix!

PR #186, opened 2019-12-25
dist: enable RW access to maddy database location

kk-boop:
This change is based on my personal experience of setting up the server. The write check kept failing until i added this configuration bit. I do note that this might just be because i messed up the setup process and had to install it in somewhat non-standard way. Also my systemd version is <235 so that might be my issue. If this does turn out to be a older systemd specific problem, could this be documented as a some sort of a known issue? Anyway i see no way of implementing this in the other service file so i do not know how to go about to fix that there as well, without hard-coding the path.
```
$ systemd --version
systemd 232
+PAM +AUDIT +SELINUX +IMA +APPARMOR +SMACK +SYSVINIT +UTMP +LIBCRYPTSETUP +GCRYPT +GNUTLS +ACL +XZ +LZ4 +SECCOMP +BLKID +ELFUTILS +KMOD +IDN

$ lsb_release -a
No LSB modules are available.
Distributor ID: Debian
Description:    Debian GNU/Linux 9.11 (stretch)
Release:        9.11
Codename:       stretch

$ uname -a
Linux machine 4.9.0-11-amd64 #1 SMP Debian 4.9.189-3 (2019-09-02) x86_64 GNU/Linux
```

### DIFF
--- a/dist/systemd/maddy.service
+++ b/dist/systemd/maddy.service
@@ -25,6 +25,7 @@ RuntimeDirectory=maddy
 StateDirectory=maddy
 LogsDirectory=maddy
 ReadOnlyPaths=/usr/lib/maddy
+ReadWritePaths=/var/lib/maddy
 
 # Strict sandboxing. You have no reason to trust code written by strangers from GitHub.
 PrivateTmp=true


foxcpp (approved):
On recent systemd StateDirectory already implies ReadWritePaths, but on 232 StateDirectory does not exist and ignored.

I do not see a reason why we cannot make it work.

foxcpp:
Thanks!

kk-boop:
No problem ;) By the "other service file" i mean the `maddy@.service`, which i thought allows for defining a non-standard location for the databases, but after looking more closely, it seems that it does not. This change can (and probably should) be thus incorporated into that service file also. Using a non-standard state directory needs changes to the service files anyway.

foxcpp:
maddy@.service allows to start multiple server instances using different configuration files. Non-standard "base directories" for various files can be in turn specified using state_dir and runtime_dir configuration directives (see [maddy(1)](https://foxcpp.dev/maddy/man/_generated_maddy.1/#global-directives)).  systemd unit is using StateDirectory, DynamicUser and other black wizardry to simplify standard installations. I am not sure there is a way to make it useful for non-standard locations. I guess word "non-standard" means users will have to figure out the remaining bits on their own.

PR #218, opened 2020-04-13
Some installation docs fixes

### LINKED ISSUE #216: libc-dev required on debian to build maddy
# Describe the bug

I noticed that building maddy with the build script fails on debian 10 if the g++ package isn't installed

# Steps to reproduce

1. Fresh VM with debian 10
2. `apt-get install gcc git curl`
3. `curl 'https://foxcpp.dev/maddy/build.sh' | bash`
This produces following error:
````
[...]
go: finding github.com/go-sql-driver/mysql v1.5.0
go: finding github.com/pierrec/lz4 v2.4.1+incompatible
# runtime/cgo
_cgo_export.c:3:10: fatal error: stdlib.h: No such file or directory
 #include <stdlib.h>
          ^~~~~~~~~~
compilation terminated.
````

Installing the `g++` package fixes this.

# Log files
[Full log](https://paste.sr.ht/%7Ebn4t/9967c914e41015bb0920e3a8ab7688c9fbcad6d5)

# Environment information
uname -a: `Linux mail 4.19.0-8-amd64 #1 SMP Debian 4.19.98-1 (2020-01-26) x86_64 GNU/Linux`

bn4t:
Changes:
- Incorporated the changes from #216 
- I also noticed that on debian the fail2ban package only recommends the python3-systemd package which is required for the maddy fail2ban jails to work. I added this too to the docs.

Let me know if something needs to be changed.

### DIFF
--- a/docs/tutorials/setting-up.md
+++ b/docs/tutorials/setting-up.md
@@ -29,10 +29,10 @@ there are a few things to keep in mind:
 Since there are currently no pre-compiled binaries for maddy, we are going to
 build it from the source. Nothing scary, this is relatively easy to do with Go.
 
-System dependencies you need to have installed is C compiler, Git and curl.
+System dependencies you need to have installed is C toolchain, Git and curl.
 On Debian-based distributions, this should be enough:
 ```
-# apt-get install gcc git curl
+# apt-get install gcc git curl libc6-dev
 ```
 
 If you want manual pages with reference docs, install scdoc too:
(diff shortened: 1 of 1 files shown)


foxcpp:
Thanks again!  I wanted to remove the fail2ban section from installation guide but now I will have to keep it :)

bn4t:
I'm curious, why did you want to remove it?

foxcpp:
In 0.2.0 there was a change to the message about failed authentication and I didn't verify whether fail2ban jails still match it properly (I adjusted the regexp used but I'm not sure if I did it correctly). I don't use fail2ban myself anywhere and it's rather tedious to verify and debug in case of problems. So... I just left it this way. There is a "hidden" disclaimer in https://github.com/foxcpp/maddy/blob/master/dist/README.md.

bn4t:
Regarding the fail2ban regex: I think it works as intended. I just tested a failed attempt and it show up as a failed one in fail2ban's log.
To be taken with a grain of salt though, I'm not very knowledgeable with fail2ban.

PR #221, opened 2020-05-08
dist: Update fail2ban filter

bn4t:
Previously the fail2ban regex would only catch failed auth messages with a smtp/submission/lmtp/imap prefix.

This regex now also catches those without such a prefix.

I tested this on my maddy server where it worked well.

### DIFF
--- a/dist/fail2ban/filter.d/maddy-auth.conf
+++ b/dist/fail2ban/filter.d/maddy-auth.conf
@@ -2,5 +2,5 @@
 before = common.conf
 
 [Definition]
-failregex    = (smtp|submission|lmtp|imap):\ authentication failed\t\{\"reason\":\".*\",\"src_ip\"\:\"<HOST>:\d+\"\,\"username\"\:\".*\"\}$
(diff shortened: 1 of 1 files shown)


foxcpp:
Missing module name prefix sounds like a bug, but quick `grep` does not reveal any messages without prefix (unless I missed something). Do you have a example when it is logged without prefix?

bn4t:
Sure. They look like this:
````
authentication failed	{"reason":"no auth. provider accepted creds, last err: unknown credentials","src_ip":"<redacted>:12542","username":"<redacted>"}
````

I can't exactly tell when they're happening. In my log I see IMAP and SMTP Submission failures logged with the correct prefix.

bn4t:
I think I found it. It seems to happen when using AUTH LOGIN. AUTH PLAINTEXT produces the correct error message.

foxcpp:
Aha, you are right. Logger is not initialized properly for SASL-related code (AUTH PLAIN is handled separately). Regarding this PR, can you adjust regexp to not check prefix at all? I think it is not necessary. P.S. The message was supposed to have prefix `submission/sasl: `.

bn4t:
Amended the commit. The regexp now no longer checks log prefixes.

foxcpp:
Thanks!

PR #227, opened 2020-05-28
Set mail from address when sending bounce emails.

wjywbs:


foxcpp:
[RFC 5321 Section 6.1](https://tools.ietf.org/html/rfc5321#section-6.1) says DSNs **MUST** be sent using null (<>) reverse path in envelope. What is the reason for standard deviation?

foxcpp:
Most importantly, null reverse path has a special meaning is that DSNs MUST NOT be generated for messages sent with it.  This prevents infinite loop of DSNs sent in response to other DSNs.

wjywbs:
Sorry for no explanation. maddy generated error "address: missing at-sign" and "Malformed sender address" when sending DSNs with empty from address. I see that other providers also send delivery failure reports using MAILER-DAEMON@autogenerated_msg_domain. I'll read the RFC.

foxcpp:
That looks like a bug in maddy. How your config looks like?

wjywbs:
```
queue outbound_queue {
    target remote {
        limits &outbound_limits
        mx_auth &outbound_auth
    }

    autogenerated_msg_domain $(primary_domain)
    bounce {
        destination some.other.domain.not.local.domains {  <- dsn fails to send
            check &outbound_checks
            modify &outbound_modifiers
            deliver_to &outbound_queue
        }
        default_destination {
            reject 550 5.0.0 "Refusing to send DSNs to non-local addresses"
        }
    }
}
```

wjywbs:
Sorry, I pasted the wrong config. I updated the previous config.

foxcpp:
Aha, so the likely source of error is this call: https://github.com/foxcpp/maddy/blob/64c91fcc7cc8d546c4de01768158e0834d77228c/internal/target/remote/remote.go#L196

The proper solution is to patch this call to handle empty return path correctly (likely by using an empty domain string for TakeMsg call below).

P.S. I hope you understand concerns outlined in https://foxcpp.dev/maddy/tutorials/alias-to-remote/ about sending bounce messages back to arbitrary addresses. This setup is very prone to creating [collateral spam](https://en.wikipedia.org/wiki/Backscatter_(email)).

wjywbs:
Yes, I'm only allowing to send bounces to pre-defined domains "some.other.domain". The default_destination still rejects to send DSNs.

foxcpp:
@wjywbs do you intend to write a patch to fix the problem described above?

wjywbs:
No, I'm not sure how to fix it.

foxcpp:
6466eaa4b1c8643ef050e1fef5cf938f4c4f82ef is the fix for your problem.

PR #228, opened 2020-05-28
Allow to sign emails from subdomains using a top domain key.

wjywbs:


foxcpp (changes requested):
Could you update man page to include newly added option?

https://github.com/foxcpp/maddy/blob/b54c705e2d3f415daa853a2e146322f4f29a8dca/docs/man/maddy-filters.5.scd#L390

docs/man/maddy-filters.5.scd
```diff
@@ -566,6 +566,11 @@ Valid values:
 Allow multiple addresses in From header field for purposes of
 require_sender_match checks. Only first address will be checked, however.
 
+*Syntax*: sign_subdomains _boolean_ ++
+*Default*: no
+
+Allow to sign emails from subdomains using a top domain key.
```
foxcpp: I think it would be useful to mention restriction to a one domain. How about...
```suggestion
Sign emails from subdomains using a top domain key.

Allows only one domain to be specified (can be workarounded using sign_dkim 
multiple times).
```

wjywbs: Thanks for the edit.

foxcpp (approved):
LGTM

foxcpp:
Thanks!

PR #230, opened 2020-05-28
Log timestamp.

wjywbs:
I'm not sure if this is wanted in this project, but this helps to track the events in the log.

### DIFF
--- a/internal/log/log.go
+++ b/internal/log/log.go
@@ -139,6 +139,11 @@ func fieldsToMap(fields []interface{}, out map[string]interface{}) {
 func (l Logger) formatMsg(msg string, fields map[string]interface{}) string {
 	formatted := strings.Builder{}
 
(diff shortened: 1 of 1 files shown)


foxcpp:
Timestamp is added by output object here: https://github.com/foxcpp/maddy/blob/e7b5c014e2e776283bed0fd2a034abaee689aac7/internal/log/writer.go#L19

By default messages are logged without it because stderr is likely to be captured by a daemon like rsyslog or systemd-journald that will add timestamps anyway. If you require timestamps for log messages written to stderr, use `stderr_ts` option for `log` directive as documented here: https://foxcpp.dev/maddy/man/_generated_maddy.1/#global-directives

foxcpp:
If you have problems with it forcing UTC timestamps - feel free to submit a PR to make it use local time, but I would like to see it including UTC offset in ISO 8601 format to keep timestamps meaningful independently of the system they are processed on.

PR #261, opened 2020-08-06
Update setting-up.md

ghost:
Replacing hostname `mx.example.org` by `example.org` accordingly to the rest of the tutorial which doesn't refer to it (i.e. DNS configuration).


foxcpp:
Do you think it might be better to use mx.example.org instead in the whole tutorial?

ghost:
Hi @foxcpp. 
Yes maybe, but we'd have to change more things in the documentation :)

foxcpp:
> Yes maybe, but we'd have to change more things in the documentation :)

Well, if it is better - why not?

ghost:
New commit. Thanks in advance for your review.

docs/tutorials/setting-up.md
```diff
@@ -110,8 +127,8 @@ own). Here is how your DNS zone should look like:
 example.org.   A     10.2.3.4
```
foxcpp, line 129: Probably these need to be updated as well to mention mx1

ghost: Done

foxcpp:
Oh well, can you resolve that conflict?

ghost:
Done.

docs/tutorials/setting-up.md
```diff
@@ -144,13 +163,14 @@ at https://mta-sts.example.org/.well-known/mta-sts.txt:
...
 
-**Note**: mx.example.org in the file is your MX hostname, example.org in URL is
-the domain you are receiving messages for. In simple configurations, they are
-may be the same, but this is not the case for more complex setups.
-If you have multiple MX servers - add them all once per line, like that:
+**Note**: mx1.example.org in the file is your MX hostname, In a simple configuration,
+it will be the same as your hostname example.org.
+In a more complex setups, you would have multiple MX servers - add them all once
+per line, like that:
+=======
```
foxcpp: Woops.

ghost: What do you mean ?

foxcpp: `======` probably should not be here.

foxcpp (approved):
According to Murphy's law, "Anything that can go wrong will go wrong".

You did not add AAAA record for MX domain and there is still a typo: "to easier" sound be "to ease".
But I'm going to merge that anyway since it is already taking way more time than it should.

PR #276, opened 2020-09-30
fix #275

### LINKED ISSUE #275: ValidMailboxName returns false for . in the mailbox part, i.e. maddy.bug@example.org
# Describe the bug

Address maddy.bug@example.org should be correct. For addresses like "maddy.bug"@example.org ValidMailboxName works fine, but it returns false if there are no quotes. However, with quotes and forwarding, recipient is not recognizes by the target servers including Maddy.

# Steps to reproduce

```
package address_test

import (
	"testing"

	"github.com/foxcpp/maddy/framework/address"
)

func TestValidMailboxName(t *testing.T) {
	if !address.ValidMailboxName("caddy.bug") {
		t.Error("caddy.bug should be valid mailbox name")
	}
}
```

# Environment information

* maddy version: master branch

acim:


foxcpp:
Thanks!

acim:
You are welcome :) Please release new image soon, I need this bug fixed there as well :)

foxcpp:
I will tag 0.4.1 and build corresponding images today (hopefully).

PR #289, opened 2020-10-30
Added port usability check in tests/t.go

hugmouse:
Because why not to


hugmouse:
In fact, this must be bad pull request. PR still has race condition. 

One solution would probably be somehow to reserve the connection in advance for future use.

foxcpp:
It is possible to bind on port 0 to get a random port. Afterwards the effective port should be available via Listener.Addr.

PR #294, opened 2020-11-22
errors refactoring

loginakhil:


internal/check/milter/milter.go
```diff
@@ -55,6 +55,7 @@ func New(_, instName string, _, inlineArgs []string) (module.Module, error) {
 	case 1:
 		c.milterUrl = inlineArgs[0]
 	case 0:
+		fallthrough
```
foxcpp: It is working as intended. See check in Init.

loginakhil: Corrected by removing the commit !

foxcpp (commented):
What is the purpose of most changes? These error values increase API surface between packages where this can be avoided and some of them can't be used in any meaningful way at all.

loginakhil:
All those errors were static ones and  as a best practise I was moving them into package level variables.
This way we don't have to call `errors.New` during every call to those functions.

foxcpp:
>All those errors were static ones and as a best practise I was moving them into package level variables.

There is no need for these variables. Not all 'best practices' should be followed unconditionally. I appreciate your enthusiasm though.

PR #307, opened 2020-11-29
Update dovecot.md to avoid cyclic dependency

reivilibre:
When I try to start Dovecot, it will complain if `/var/run/maddy/` is missing as it can't create /var/run/maddy/dovecot-lmtp.sock`.
When I try to start Maddy, it complains that `unix:///var/run/dovecot/auth-maddy-client` doesn't exist.

Therefore I can't start either service.
I decided to move the LMTP socket into Dovecot's runtime directory since it provides the socket,
and to remove the sock extension to match the naming convention in the rest of the directory.

### DIFF
--- a/docs/third-party/dovecot.md
+++ b/docs/third-party/dovecot.md
@@ -17,7 +17,7 @@ protocols = imap lmtp
 
 # /etc/dovecot/conf.d/10-master.conf
 service lmtp {
- unix_listener /var/run/maddy/dovecot-lmtp.sock {
+ unix_listener lmtp-maddy {
    mode = 0600
    user = maddy
   }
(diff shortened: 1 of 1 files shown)


foxcpp:
Wouldn't it cause `Permission denied` errors if [systemd unit file with sandboxing](https://github.com/foxcpp/maddy/blob/master/dist/systemd/maddy.service) is used? It has `ProtectSystem=strict` that makes everything except for whitelisted directories read-only. Perhaps the documentation should mention the need to add `ReadWritePaths=/var/lib/dovecot` then

reivilibre:
I haven't ran into that yet; I am using the systemd unit file with sandboxing.
However, that would also affect the SASL auth socket.

I can't guarantee this isn't a problem yet because my mailserver isn't actually functional yet...

foxcpp:
Do you have any updates on that? Is your setup now functional?

reivilibre:
@foxcpp Yup, my set-up is working! Lovely piece of kit, thank you very much!

foxcpp:
Okay, let's get this in then. Thanks!

PR #309, opened 2020-11-30
Suggest disabling TLS for Dovecot LMTP unix socket

reivilibre:
I don't see a point to TLS over a UNIX socket anyway, and it caused me trouble.


foxcpp:
Thanks!

PR #319, opened 2020-12-06
Copyrights fix

v1a0:


foxcpp:
NACK. This copyright example is part of the license document that cannot be changed:
> Everyone is permitted to copy and distribute verbatim copies
 of this license document, but changing it is not allowed.

v1a0:
I'm sorry my fault

PR #320, opened 2020-12-15
Patch Dockerfile for v0.4.3

natsukagami:
- Update stage 2 base image to `alpine:3.12` to match stage 1
- Use new syntax for `./build.sh`
- Patch the default `state_dir` and `runtime_dir` in the default `maddy.conf` so that it's backwards-compatible with v0.4.2.
  - This is not enough I think, I would still need to update my config file if I already wrote one (which I expect everyone to!)

Should I update the script so that it injects the old `state_dir` and `runtime_dir` in?

---
This change is [Reviewable](https://reviewable.io/reviews/foxcpp/maddy/320)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ COPY . maddy/
 WORKDIR maddy/
 
 ENV LDFLAGS -static
-RUN apk --no-cache add bash git gcc musl-dev
+RUN apk --no-cache add bash git gcc musl-dev scdoc
 
 RUN mkdir /pkg/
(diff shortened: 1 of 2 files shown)


foxcpp:
Thanks for your work but I took a slightly different approach:
```
+RUN echo -e "state_dir /data\nruntime_dir /tmp" >> /pkg/data/maddy.conf
```

Adding state_dir/runtime_dir to default configuration will make it Linux-specific.
Injecting necessary directives with Docker-specific values while building the image seems to be a better idea.

PR #332, opened 2021-01-16
check: Add missing arg in skeleton impl of CheckState

hhirtz:
Module skeleton in internal/check was missing the context argument defined here:

<https://github.com/foxcpp/maddy/blob/baba1d9c590313a6b347ec215318d32d2b24f919/framework/module/check.go#L59>


foxcpp:
Thanks!

PR #343, opened 2021-03-15
config/tls: Fix tls_client parsing

AluisioASG:
`tls_client` recurses into processing itself and causes a stack overflow when it's specified in a configuration file, at least during my evaluation of maddy. I looked into what server.go does to fix it.

This also replaces the conditional for loading a client keypair with a more straightforward one. I suspect it may have been intentional, but as it's not documented I'm leaning towards it being a mistake.

I've verified that `tls_client` still works inside a `target.remote` block, and presents the configured certificate when requested.


foxcpp:
Thanks!

PR #345, opened 2021-03-22
Fix a typo in maddy-smtp man doc

Defman21:
I assume it was a leftover from copy-pasting stuff because I couldn't find a directive named `destination_to` anywhere.


foxcpp:
Thanks!

Issue #324, opened 2021-01-06
endpoint/smtp: Reimplement header size limit


foxcpp:
textproto package in go-message 0.14.1 no longer restricts the header size. It should be implemented in maddy.

Ref. https://github.com/emersion/go-message/releases/tag/v0.14.1

foxcpp:
Oh. I accidentally upgraded go-message to 0.14.1 in 0.4.4 so this is a regression.

foxcpp:
With the mentioned regression and default configuration, maddy could consume up to 1280 MiB of RAM while processing inbound messages. This is a DoS risk factor.

PR #362, opened 2021-06-20
Limit allowed header size

apreiml:
Implements header file check mentioned here https://github.com/foxcpp/maddy/issues/324

I've also added a test for max_message_size as subtest in the new test commited here. It fails with an internal server error tough. I'll check what's happening there and add the test in a separate pull request.

### DIFF
--- a/internal/endpoint/smtp/session.go
+++ b/internal/endpoint/smtp/session.go
@@ -40,6 +40,31 @@ import (
 	"github.com/foxcpp/maddy/framework/module"
 )
 
+func limitReader(r io.Reader, n int64, err error) *limitedReader {
+	return &limitedReader{R: r, N: n, E: err, Enabled: true}
+}
+
+type limitedReader struct {
+	R       io.Reader
+	N       int64
+	E       error
+	Enabled bool
+}
+
+// same as io.LimitedReader.Read except returning the custom error and the option
+// to be disabled
+func (l *limitedReader) Read(p []byte) (n int, err error) {
+	if l.Enabled && l.N <= 0 {
+		return 0, l.E
+	}
+	if int64(len(p)) > l.N {
+		p = p[0:l.N]
+	}
+	n, err = l.R.Read(p)
+	l.N -= int64(n)
+	return
+}
+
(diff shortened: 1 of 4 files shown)


apreiml:
Does not work as intended

apreiml:
Ok, now it doesn't apply the header file size restriction to the rest of the message.

If this implementation is not ok, another Idea may be to disable the limit check after reading the header. Because afaik the message size will then be handled by go-smtp

foxcpp:
>If this implementation is not ok, another Idea may be to disable the limit check after reading the header. Because afaik the message size will then be handled by go-smtp

That sounds a bit more cleaner. 

Also, I believe bufio.Reader may read up to 4096 extra bytes from the underlying reader. Not sure if there is a need to account for this somehow.

internal/endpoint/smtp/session.go
```diff
@@ -353,6 +389,12 @@ func (s *Session) prepareBody(ctx context.Context, r io.Reader) (textproto.Heade
 		}
 	}
 
+	limitr.AddLimit(s.endp.serv.MaxMessageBytes-s.endp.maxHeaderBytes, &exterrors.SMTPError{
+		Code:         552,
+		EnhancedCode: exterrors.EnhancedCode{5, 3, 4},
+		Message:      "Max message size exceeded",
+	})
+
 	buf, err := s.endp.buffer(bufr)
```
apreiml, line 390: @foxcpp I guess the current option is better. Otherwise if I disable the limit before this line, the whole message will be put into the buffer. Having a limit like this, will make the buffering fail now with the proper error.

apreiml:
Ok, now the limitedReader is disabled after reading the header.

> Also, I believe bufio.Reader may read up to 4096 extra bytes from the underlying reader. Not sure if there is a need to account for this somehow.

Yes, this is true. An option would be to add the .Size() of the buffio.Reader to the limitedReader. Then the buffer would have read at least the header size before failing.

foxcpp (approved):
LGTM

foxcpp:
Thanks!

PR #365, opened 2021-07-02
Add missing mailto: to DMARC ruf parameter in tutorial

arisudesu:
Looks like it must include `mailto:` scheme to be valid. The rfc7489 does not mention it explicitly, though, I find no examples in the web for using it without scheme. Especially examples on dmarc.org itself.


foxcpp:
Good catch. Thanks!

PR #378, opened 2021-07-31
refactor: remove/`_`-ify unused params

Gusted:
Hi!

I've removed some unused params. But if they where needed for e.g. interface type I've simply `_` them. Also I have 2 instances to fix tests params, whereby they were passed but not initialized at all, they are in`internal/target/remote/remote_test.go` and `internal/modify/dkim/dkim_test.go`. All test are still passing so it seems like I didn't break anything.

I might've refactored some code away that actually is used but wasn't implemented correctly, but as far as I see their is nothing wrong or erroring going on.

### DIFF
--- a/config.go
+++ b/config.go
@@ -40,7 +40,7 @@ type logOut struct {
 	log.Output
 }
 
-func logOutput(m *config.Map, node config.Node) (interface{}, error) {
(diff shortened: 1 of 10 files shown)


foxcpp:
Looks like most changes are harmless. Thanks!

PR #380, opened 2021-08-01
refactor: left-overs

Gusted:
Apparently my previous PR didn't cover everything. Also some errors where hidden from me so might see some more new refactor cases here.

### DIFF
--- a/config.go
+++ b/config.go
@@ -82,7 +82,7 @@ func LogOutputOption(args []string) (log.Output, error) {
 			// keep the absolute path for reinitialization.
 			args[i] = absPath
 
-			w, err := os.OpenFile(absPath, os.O_WRONLY|os.O_APPEND|os.O_CREATE, 0666)
+			w, err := os.OpenFile(absPath, os.O_WRONLY|os.O_APPEND|os.O_CREATE, 0o666)
 			if err != nil {
 				return nil, fmt.Errorf("failed to create log file: %v", err)
 			}
--- a/framework/config/module/check_action.go
+++ b/framework/config/module/check_action.go
@@ -56,7 +56,7 @@ type FailAction struct {
 	ReasonOverride *exterrors.SMTPError
 }
 
-func FailActionDirective(m *config.Map, node config.Node) (interface{}, error) {
+func FailActionDirective(_ *config.Map, node config.Node) (interface{}, error) {
 	if len(node.Children) != 0 {
 		return nil, config.NodeErr(node, "can't declare block here")
 	}
--- a/framework/config/tls/server.go
+++ b/framework/config/tls/server.go
@@ -88,7 +88,7 @@ func readTLSBlock(globals map[string]interface{}, blockNode config.Node) (*TLSCo
 
 	childM.Custom("loader", false, false, func() (interface{}, error) {
 		return loader, nil
-	}, func(m *config.Map, node config.Node) (interface{}, error) {
+	}, func(_ *config.Map, node config.Node) (interface{}, error) {
 		var l module.TLSLoader
 		err := modconfig.ModuleFromNode("tls.loader", node.Args, node, globals, &l)
 		return l, err
--- a/framework/logparser/parse_test.go
+++ b/framework/logparser/parse_test.go
@@ -50,19 +50,19 @@ func TestParse(t *testing.T) {
 	}
 
 	test("2006-01-02T15:04:05.000Z module: hello\t", Msg{
-		Stamp:   time.Date(2006, time.January, 02, 15, 04, 05, 0, time.UTC),
+		Stamp:   time.Date(2006, time.January, 2, 15, 4, 5, 0, time.UTC),
 		Module:  "module",
 		Message: "hello",
 		Context: map[string]interface{}{},
 	}, "")
 	test("2006-01-02T15:04:05.000Z module: hello: whatever\t", Msg{
-		Stamp:   time.Date(2006, time.January, 02, 15, 04, 05, 0, time.UTC),
+		Stamp:   time.Date(2006, time.January, 2, 15, 4, 5, 0, time.UTC),
 		Module:  "module",
 		Message: "hello: whatever",
 		Context: map[string]interface{}{},
 	}, "")
 	test("2006-01-02T15:04:05.000Z module: hello: whatever\t{\"a\":1}", Msg{
-		Stamp:   time.Date(2006, time.January, 02, 15, 04, 05, 0, time.UTC),
+		Stamp:   time.Date(2006, time.January, 2, 15, 4, 5, 0, time.UTC),
 		Module:  "module",
 		Message: "hello: whatever",
 		Context: map[string]interface{}{
@@ -70,7 +70,7 @@ func TestParse(t *testing.T) {
 		},
 	}, "")
 	test("2006-01-02T15:04:05.000Z module: hello: whatever\t{\"a\":1,\"b\":\"bbb\"}", Msg{
-		Stamp:   time.Date(2006, time.January, 02, 15, 04, 05, 0, time.UTC),
+		Stamp:   time.Date(2006, time.January, 2, 15, 4, 5, 0, time.UTC),
 		Module:  "module",
 		Message: "hello: whatever",
 		Context: map[string]interface{}{
@@ -79,7 +79,7 @@ func TestParse(t *testing.T) {
 		},
 	}, "")
 	test("2006-01-02T15:04:05.000Z [debug] module: hello: whatever\t{\"a\":1,\"b\":\"bbb\"}", Msg{
-		Stamp:   time.Date(2006, time.January, 02, 15, 04, 05, 0, time.UTC),
+		Stamp:   time.Date(2006, time.January, 2, 15, 4, 5, 0, time.UTC),
 		Debug:   true,
 		Module:  "module",
 		Message: "hello: whatever",
@@ -89,7 +89,7 @@ func TestParse(t *testing.T) {
 		},
 	}, "")
 	test("2006-01-02T15:04:05.000Z [debug] oink oink: hello: whatever\t{\"a\":1,\"b\":\"bbb\"}", Msg{
-		Stamp:   time.Date(2006, time.January, 02, 15, 04, 05, 0, time.UTC),
+		Stamp:   time.Date(2006, time.January, 2, 15, 4, 5, 0, time.UTC),
 		Debug:   true,
 		Message: "oink oink: hello: whatever",
 		Context: map[string]interface{}{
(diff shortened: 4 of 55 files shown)


framework/config/lexer/lexer.go
```diff
@@ -113,7 +114,7 @@ func (l *lexer) next() bool {
 			if len(val) > 0 {
 				return makeToken()
 			}
-			if err == io.EOF {
+			if errors.Is(err, io.EOF) {
```
foxcpp: According to https://github.com/golang/go/issues/39155 this use of `errors.Is` is not useful. I agree with reasoning in https://github.com/golang/go/issues/39155#issuecomment-631444428 that `io.EOF` should be seen as a special value rather than an error and thus should not be used with `errors.Is`.

framework/config/tls/server.go
```diff
@@ -96,7 +96,7 @@ func readTLSBlock(globals map[string]interface{}, blockNode config.Node) (*TLSCo
 
 	childM.Custom("loader", false, false, func() (interface{}, error) {
 		return loader, nil
-	}, func(m *config.Map, node config.Node) (interface{}, error) {
+	}, func(_ *config.Map, _ config.Node) (interface{}, error) {
 		var l module.TLSLoader
 		err := modconfig.ModuleFromNode("tls.loader", blockNode.Args, config.Node{}, globals, &l)
```
foxcpp: This code looks suspicious. Shouldn't it be
```go
err := modconfig.ModuleFromNode("tls.loader", node.Args, node, globals, &l)
```
aka processing config block for `loader` directive.

framework/logparser/parse_test.go
```diff
@@ -50,27 +50,27 @@ func TestParse(t *testing.T) {
 	}
 
 	test("2006-01-02T15:04:05.000Z module: hello\t", Msg{
-		Stamp:   time.Date(2006, time.January, 02, 15, 04, 05, 0, time.UTC),
+		Stamp:   time.Date(2006, time.January, 0o2, 15, 0o4, 0o5, 0, time.UTC),
```
foxcpp: That seems to be the unintended misuse of octal literals and probably it should be `2, 15, 4, 5` instead (without leading zeros). Same for other cases in this file.

internal/auth/shadow/module.go
```diff
@@ -91,11 +91,11 @@ func (a *Auth) Lookup(username string) (string, bool, error) {
 
 	ent, err := Lookup(username)
 	if err != nil {
-		return "", false, nil
+		return "", false, err
 	}
 
 	if !ent.IsAccountValid() {
-		return "", false, nil
+		return "", false, fmt.Errorf("shadow: account is expired")
```
foxcpp: `Lookup` result in this case indicates whether the account exists and can be used. In general errors returned by `Lookup` indicate that result is not meaningful (in this case - we **don't know** whether the account exists). I would like to keep the current semantics - pretend expired accounts do not exist instead of hard-failing lookup.

internal/auth/shadow/module.go
```diff
@@ -91,11 +91,11 @@ func (a *Auth) Lookup(username string) (string, bool, error) {
 
 	ent, err := Lookup(username)
 	if err != nil {
-		return "", false, nil
+		return "", false, err
```
foxcpp, line 97: I agree with this change but `ErrNoSuchUser` should be special-cased to return `nil` here indicating a non-existing account.

internal/endpoint/smtp/smtp.go
```diff
@@ -156,7 +157,7 @@ func autoBufferMode(maxSize int, dir string) func(io.Reader) (buffer.Buffer, err
 				log.Debugln("autobuffer: keeping the message in RAM (read", actualSize, "bytes, got EOF)")
 				return buffer.MemoryBuffer{Slice: initial[:actualSize]}, nil
 			}
-			if err == io.EOF {
+			if errors.Is(err, io.EOF) {
```
foxcpp: Ditto.

internal/modify/dkim/dkim.go
```diff
@@ -327,12 +327,12 @@ func (s *state) RewriteBody(ctx context.Context, h *textproto.Header, body buffe
 		var err error
 		domain, err = idna.ToASCII(domain)
 		if err != nil {
-			return nil
+			return err
```
foxcpp: Needs discussion on what the desired behavior should be.

1. Hard-failing message delivery if it cannot be signed due to use of IDNA names in DKIM identifiers.
  (there is no reason anyone would do that in practice but it is possible and RFCs explicitly mention that possibility so we have to handle it)
2. Sending the message without the signature.

2 is how it was before, you changed it into 1.

config.go
```diff
@@ -62,7 +62,7 @@ func LogOutputOption(args []string) (log.Output, error) {
 		case "syslog":
 			syslogOut, err := log.SyslogOutput()
 			if err != nil {
-				return nil, fmt.Errorf("failed to connect to syslog daemon: %v", err)
+				return nil, fmt.Errorf("failed to connect to syslog daemon: %w", err)
```
foxcpp: Are you sure `%w` is useful in this case? This increases API surface of LogOutputOption and breaks the abstraction.

See https://blog.golang.org/go1.13-errors#TOC_3.4 for `%v` vs `%w` recommendations provided by Go team.

framework/config/module/check_action.go
```diff
@@ -140,7 +140,7 @@ func ParseRejectDirective(args []string) (*exterrors.SMTPError, error) {
 	case 1:
 		code, err = strconv.Atoi(args[0])
 		if err != nil {
-			return nil, fmt.Errorf("invalid error code integer: %v", err)
+			return nil, fmt.Errorf("invalid error code integer: %w", err)
```
foxcpp: Ditto. Unjustified use of `%w`.

internal/auth/dovecot_sasl/dovecot_sasl.go
```diff
@@ -99,7 +99,7 @@ func (a *Auth) Init(cfg *config.Map) error {
 
 	endp, err := config.ParseEndpoint(a.serverEndpoint)
 	if err != nil {
-		return fmt.Errorf("%s: invalid server endpoint: %v", modName, err)
+		return fmt.Errorf("%s: invalid server endpoint: %w", modName, err)
```
foxcpp: Unjustified use of `%w`?

internal/auth/dovecot_sasl/dovecot_sasl.go
```diff
@@ -108,12 +108,12 @@ func (a *Auth) Init(cfg *config.Map) error {
 	// Dial once to check usability and also to get list of mechanisms.
 	conn, err := net.Dial(endp.Scheme, endp.Address())
 	if err != nil {
-		return fmt.Errorf("%s: unable to contact server: %v", modName, err)
+		return fmt.Errorf("%s: unable to contact server: %w", modName, err)
```
foxcpp: Ditto.

internal/auth/dovecot_sasl/dovecot_sasl.go
```diff
@@ -108,12 +108,12 @@ func (a *Auth) Init(cfg *config.Map) error {
...
 	}
 
 	cl, err := dovecotsasl.NewClient(conn)
 	if err != nil {
-		return fmt.Errorf("%s: unable to contact server: %v", modName, err)
+		return fmt.Errorf("%s: unable to contact server: %w", modName, err)
```
foxcpp: Ditto.

internal/auth/shadow/module.go
```diff
@@ -76,7 +76,7 @@ func (a *Auth) Init(cfg *config.Map) error {
 			if os.IsPermission(err) {
 				return fmt.Errorf("shadow: can't read /etc/shadow due to permission error, use helper binary or run maddy as a privileged user")
 			}
-			return fmt.Errorf("shadow: can't read /etc/shadow: %v", err)
+			return fmt.Errorf("shadow: can't read /etc/shadow: %w", err)
```
foxcpp: Unjustified use of `%w`?

foxcpp (changes requested):
This PR mixes minor style fixes, bug fixes and changes that change server behavior in some edge cases. I would appreciate if you group these changes into separate commits.

Primarily I'm concerned about use of %w everywhere which is something Go developers seem to advise against.

framework/config/lexer/lexer.go
Gusted: I din't knew about that discussion/proposal. Thanks for linking!

internal/modify/dkim/dkim.go
Gusted: > RFCs explicitly mention that possibility so we have to handle it

Could you link that? I'm just seeing that rfc5890 is mentioned that the IDNA is explicitly is a UTF-8 encoded string.

config.go
Gusted: Hmm that's a valid point. From my understanding `%w` will output the unwrapped error as a whole. Which is nice if some error is coming up and you can exactly see which error is the "root" cause. And thus I personally am using that. However if it breaks the abstraction I will happily revert it.

Gusted:
The PR should now have multiple commits.

config.go
foxcpp: "full error chain" is already visible in error messages formatted using %v, %w makes the chain accessible to the code and this means you are commiting to returning same wrapped errors in the future - the paragraph i linked from go blog shows how it may make it harder to change code later if used carelessly.

internal/modify/dkim/dkim.go
foxcpp: https://datatracker.ietf.org/doc/html/rfc8616#section-5

Gusted: Hmm, I think we should better hold it for situation 2, and in another PR/issue change it to something like situation 1.

foxcpp:
Holding this off until 0.5 release that I will make soon.

foxcpp:
Included fix for TLS config in e025d3b11c4e9f4751c90fe683a9d0a18cd94675.

internal/storage/imapsql/imapsql.go
```diff
@@ -478,12 +478,12 @@ func (store *Storage) GetOrCreateIMAPAcct(username string) (backend.User, error)
 func (store *Storage) Lookup(key string) (string, bool, error) {
 	accountName, err := prepareUsername(key)
 	if err != nil {
-		return "", false, nil
+		return "", false, err
```
foxcpp: I believe it should stay `nil` for the same reason as other Lookup implementations you changed.

Gusted: Got it, done.

Gusted:
Code is rebased on master and still passes all tests.

foxcpp (approved):
LGTM

foxcpp:
Thanks!

PR #391, opened 2021-08-22
Set ExecStart path to the same as the default in build.sh

### LINKED ISSUE #390: ExecStart path is incorrect
# Describe the bug

The `ExecStart` path in the systemd startup script is incorrect by default.
It is set to `/usr/bin/maddy` but maddy gets installed to `/usr/local/bin/maddy` when installing via the build script.

# Environment information

* maddy version: 0.5.0 and master

jpatters:
This is really only a half fix as I see that you can pass a prefix to the build script. 
Closes #390


foxcpp:
Thanks!

PR #394, opened 2021-08-26
dist: Set ProtectClock, RestrictAddressFamilies and CapabilityBoundingSet

erdnaxe:
This increases the isolation of Maddy service.
Maddy capabilities can be bound to only `CAP_NET_BIND_SERVICE` with `CapabilityBoundingSet`.
`RestrictAddressFamilies` restricts the service to only use Unix sockets, IPv4 and IPv6.

-- 
Alexandre
Thank you for developing this awesome mail server.


foxcpp:
Thanks!

PR #414, opened 2021-10-13
smtp/session: move conditional to outer loop

Gusted:
- Only loop trough the `from` email address if UTF8 option is not enabled. This should avoid unnecessary loops for servers which has the UTF8 option enabled.


foxcpp:
Thanks!

PR #416, opened 2021-11-17
dmarc: only ResultPass if both DKIM and SPF are Aligned

mschneider82:
In my case the DKIM Result was fine, but SPF fails, In that case i expect an DMARC failure. But the Dmarc was ResultNone

Please correct me if i am wrong.


mschneider82:
it seems that the `or` was correct!

https://support.google.com/a/answer/10032169#alignment

Issue #404, opened 2021-09-19
Docker support for ARM based machines


vedangwartikar:
# Use case

Currently, the Docker image supports building maddy for AMD machines. 
Additional support for aarch64/arm architecture-based machines.

# Your idea for a solution

Create another Docker image for arm-based machines

- [x] I'm willing to help with the implementation

foxcpp:
There is nothing preventing the building of maddy image for ARM. In fact I did so in the past. But I do not know how to build and upload both amd and arm images to DockerHub.

vedangwartikar:
When I'm trying to set up maddy from https://hub.docker.com/r/foxcpp/maddy using the docker run command on my aarch64 based VM, I'm getting the error

`standard_init_linux.go:211: exec user process caused "exec format error"`

Looked it up and found out that it might be related to the architecture. Not sure if I'm doing anything wrong here.

foxcpp:
DockerHub only has the amd64 image so this is not going to work currently.

foxcpp:
You should be able to build the image yourself and run it, though.
```
git clone https://github.com/foxcpp/maddy.git
docker bulld -t foxcpp/maddy:0.5.1
```

maitredede:
Hello,
I am working on golang projects with docker images that runs multi-arch. I have successfully built multi-arch images using the new [buildx](https://github.com/docker/buildx) plugin for docker, and a raspberry pi (I had troubles with golang running on qemu, so I choosed native hardware).
It is quite easy to set up : on the pi, use raspbian arm64, install docker server, and expose it (to your LAN). Then on the amd64 machine (running also docker server), with buildx installed, you can create runnners targetting these docker servers, with explicit arch configuration. And the build/push is launched with one simple commandline.
I will try to make a doc for this.

AlphaJack:
@foxcpp any update on #491 implementation?  
Otherwise [filebrowser](https://github.com/filebrowser/filebrowser/blob/master/.goreleaser.yml) is using goreleaser, which seems trivial to configure

foxcpp:
0.6 will have an arm64 image available. Will add 32-bit ARM if someone specifically requests it.

AlphaJack:
@foxcpp do you have an approximate ETA on the next release?

foxcpp:
@AlphaJack in the next few days

foxcpp:
[...]

PR #418, opened 2021-11-23
multi-arch build starting point

maitredede:
As stated in #404, here is a starting point to build easily multi-arch images.


foxcpp:
Thanks!

PR #449, opened 2022-01-31
Add more fields to imap_filter

angelnu:
I added the missing rcpts and address as in the check commands. This helps me to filter based on the original recipe of the email which I expect to change less than the sender.

### DIFF
--- a/internal/imap_filter/command/command.go
+++ b/internal/imap_filter/command/command.go
@@ -28,6 +28,7 @@ import (
 	"os"
 	"os/exec"
 	"regexp"
+	"strings"
 
 	"github.com/emersion/go-message/textproto"
 	"github.com/foxcpp/maddy/framework/buffer"
@@ -127,15 +128,23 @@ func (c *Check) expandCommand(msgMeta *module.MsgMetadata, accountName string) (
 				valI, err := msgMeta.Conn.RDNSName.Get()
 				if err != nil {
 					return ""
-				}
+				}x
 				if valI == nil {
-					return ""
+					return ""x
 				}
 				return valI.(string)
 			case "{msg_id}":
 				return msgMeta.ID
 			case "{sender}":
 				return msgMeta.OriginalFrom
+			case "{rcpts}":
+				rcpts := []string{}
+				for _, value := range msgMeta.OriginalRcpts {
+					rcpts = append(rcpts, value)
+				}
+				return strings.Join(rcpts, "\n")
+			case "{address}":
+				return msgMeta.OriginalRcpts[accountName]
 			case "{account_name}":
 				return accountName
 			}
--- a/internal/imap_filter/command/command.go
+++ b/internal/imap_filter/command/command.go
@@ -49,8 +49,8 @@ type Check struct {
 	cmdArgs []string
 }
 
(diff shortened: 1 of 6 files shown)


angelnu:
@foxcpp - I copied the code for rcpts and therefore it uses new-line separators. Maybe we prefer commas as for the aliases?

internal/imap_filter/command/command.go
```diff
@@ -127,15 +128,23 @@ func (c *Check) expandCommand(msgMeta *module.MsgMetadata, accountName string) (
...
 				if valI == nil {
-					return ""
+					return ""x
 				}
 				return valI.(string)
 			case "{msg_id}":
 				return msgMeta.ID
 			case "{sender}":
 				return msgMeta.OriginalFrom
+			case "{rcpts}":
```
foxcpp: I wonder if it is a better idea to name this original_rcpts to avoid confusion.

internal/imap_filter/command/command.go
```diff
@@ -127,15 +128,23 @@ func (c *Check) expandCommand(msgMeta *module.MsgMetadata, accountName string) (
...
+				rcpts := []string{}
+				for _, value := range msgMeta.OriginalRcpts {
+					rcpts = append(rcpts, value)
+				}
+				return strings.Join(rcpts, "\n")
+			case "{address}":
+				return msgMeta.OriginalRcpts[accountName]
```
foxcpp: accountName is not guaranteed to be a email address at all.
E.g. https://maddy.email/multiple-domains/

IMAPFilter interface does not provide access to SMTP envelope information. I wonder how it should be (if should) extended to do.

angelnu: While I was testing I noticed that even when I send the email to multiple recipients rcpts would only have one single address. I do not not know if this is a bug in either my test or Maddy but it makes address to not being required.

On the other hand I also got one without  OriginalRcpts as the email was sent to my real email address (not a domain alias).

So it looks like having access to the original TO address is needed to make this more reliable...

internal/imap_filter/command/command.go
angelnu: please see my reply bellow - it looks like OriginalRcpts is not set when the Rcpts is not replaced. Will need to fix that. Then I will rename to originalRcpts as proposed.

angelnu: renamed. Also ensured all original rcpts are added and not only those modified by the msgpipeline

internal/imap_filter/command/command.go
angelnu: made the changes to pass the rctp_to

internal/storage/imapsql/delivery.go
```diff
@@ -32,13 +32,16 @@ import (
 	"github.com/foxcpp/maddy/internal/target"
 )
 
+type addecRcpt struct {
```
foxcpp: Typo? addedRcpt

angelnu: indeed - fixed

foxcpp:
Manually merged into dev branch as https://github.com/foxcpp/maddy/commit/d60ce8321f1b0c030e6e54671e18d9582d23973d. Thanks a lot for your work!

PR #450, opened 2022-02-13
bugfix: invert path check

jonlundy:
Hi i found this bug when trying to use dovecot_sasl on a remote host.


foxcpp:
Woops, thanks!

PR #455, opened 2022-02-20
test: use `T.TempDir` to create temporary test directory

Juneezee:
A small testing enhancement here.

We can use the `T.TempDir` function from the `testing` package to create temporary directory. The directory created by `T.TempDir` is automatically removed when the test and all its subtests complete.

Reference: https://pkg.go.dev/testing#T.TempDir

### DIFF
--- a/internal/target/queue/queue_test.go
+++ b/internal/target/queue/queue_test.go
@@ -45,20 +45,13 @@ import (
 // See newTestQueueDir to create testing queue from an existing directory.
 // It is called responsibility to remove queue directory created by this function.
 func newTestQueue(t *testing.T, target module.DeliveryTarget) *Queue {
-	dir, err := ioutil.TempDir("", "maddy-tests-queue")
-	if err != nil {
-		t.Fatal("failed to create temporary directory for queue:", err)
-	}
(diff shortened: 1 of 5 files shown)


internal/target/queue/queue_test.go
```diff
@@ -45,11 +45,7 @@ import (
 // See newTestQueueDir to create testing queue from an existing directory.
 // It is called responsibility to remove queue directory created by this function.
 func newTestQueue(t *testing.T, target module.DeliveryTarget) *Queue {
-	dir, err := ioutil.TempDir("", "maddy-tests-queue")
-	if err != nil {
-		t.Fatal("failed to create temporary directory for queue:", err)
-	}
-	return newTestQueueDir(t, target, dir)
+	return newTestQueueDir(t, target, t.TempDir())
```
foxcpp, line 48: os.RemoveAll call in cleanQueue function below seems redundant now. Can you remove it?

Juneezee: Done :heavy_check_mark:

foxcpp (approved):

foxcpp:
Thanks!

PR #478, opened 2022-04-04
dmarc: fixed pct calculation

mschneider82:


mschneider82:
oh sorry, the logic was correct! it is negated

PR #516, opened 2022-07-02
Remove references to man.5 from build.sh

MFAshby:
man.5 page was removed in c0eacfa0f34f4040ec4fd5d3ca44a35e8cde21b4
removed remaining references in build.sh, otherwise strict build (like arch linux makepkg) fails.


foxcpp:
Thanks!

PR #532, opened 2022-09-06
internal/auth: Add native NetAuth authentication integration.

the-maldridge:
NetAuth is an easy to deploy and manage organizational source of truth for identity and group management.  This commit adds support for maddy to ask NetAuth about authentication using the native protocol.

There are problems that I haven't figured out yet, mostly that config parsing doesn't seem to actually work in my patch, but the auth works fine.  NetAuth expects usernames to be *nix compatible so it pretends to accept emails and then strips the domain component off.  This appeared to be the most straightforward way to make this integration work.

I think ultimately I'd like to hide this behind a build flag, but for now I'm opening this to get feedback for why the config parsing doesn't work.


internal/auth/netauth/netauth.go
```diff
@@ -0,0 +1,128 @@
...
+	a.log.Println("In netauth init")
+	l := hclog.New(&hclog.LoggerOptions{Output: a.log})
+	n, err := netauth.NewWithLog(l)
+	if err != nil {
+		return err
+	}
+	a.nacl = n
+	a.nacl.SetServiceName("maddy")
+	cfg.String("require_group", false, false, "", &a.mustGroup)
+	cfg.Bool("debug", true, false, &a.log.Debug)
```
foxcpp, line 51: You need a cfg.Process call after defining all configuration directives.

the-maldridge: Hmm, I dunno how I missed that from the ldap file I was referring to, but I have added the call.  I will try to re-test soon.

internal/auth/netauth/netauth.go
```diff
@@ -0,0 +1,132 @@
...
+		return "", false, err
+	}
+
+	e, err := a.nacl.EntityInfo(ctx, localpart)
+	if err != nil {
+		return "", false, fmt.Errorf("%s: search: %w", modName, err)
+	}
+
+	if a.mustGroup != "" {
+		a.log.Msg("Performing require_group check")
```
foxcpp: That probably should be a debug-level message.

internal/auth/netauth/netauth.go
```diff
@@ -0,0 +1,132 @@
...
+// plugin.  Given the way that NetAuth works it doesn't really make
+// sense to have more than one instance, but this is part of the API.
+func (a *Auth) InstanceName() string {
+	return a.instName
+}
+
+// Lookup requests the entity from the remote NetAuth server,
+// potentially returning that the user does not exist at all.
+func (a *Auth) Lookup(ctx context.Context, username string) (string, bool, error) {
+	localpart, _, err := address.Split(username)
```
foxcpp: Probably needs to be removed, how such authentication providers are intended to be used is described in https://maddy.email/tutorials/pam/#account-names

foxcpp (changes requested):

internal/auth/netauth/netauth.go
the-maldridge: yep, a leftover from before I had config parsing actually doing anything.

internal/auth/netauth/netauth.go
the-maldridge: Changed per directions.  Those documents explain how to do this with the imap storage, but submission also references the auth block.  Can you add some to the docs about how to ensure that that also does the translation?

foxcpp: storage.imapsql is used by both delivery (submission) and IMAP.  delivery_map is applied specifically for delivery part. 

The IMAP part is supposed to be handled by auth_map, but that's not necessary in PAM/netauth cases - one could just use usernames without domain part as a login. Hence, it is not used in the tutorial.

internal/auth/netauth/netauth.go
```diff
@@ -0,0 +1,121 @@
...
+func New(modName, instName string, _, inlineArgs []string) (module.Module, error) {
+	return &Auth{
+		instName: instName,
+		log:      log.Logger{Name: modName},
+	}, nil
+}
+
+// Init performs deferred initialization actions.
+func (a *Auth) Init(cfg *config.Map) error {
+	a.log.Println("In netauth init")
```
foxcpp: Another left-over from a debugging session?

internal/auth/netauth/netauth.go
```diff
@@ -0,0 +1,121 @@
...
+// AuthPlain attempts straightforward authentication of the entity on
+// the remote NetAuth server.
+func (a *Auth) AuthPlain(username, password string) error {
+	a.log.Debugf("attempting to auth user: %s", username)
+	if err := a.nacl.AuthEntity(context.Background(), username, password); err != nil {
+		return module.ErrUnknownCredentials
+	}
+	a.log.Debugln("netauth returns successful auth")
+	if a.mustGroup != "" {
+		a.log.Msg("Performing require_group check")
```
foxcpp: Another left-over from a debugging session?

internal/auth/netauth/netauth.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func (a *Auth) checkMustGroup(username string) error {
+	groups, err := a.nacl.EntityGroups(context.Background(), username)
+	if err != nil {
+		return fmt.Errorf("%s: groups: %w", modName, err)
+	}
+	found := false
+	for _, g := range groups {
+		found = found || (g.GetName() == a.mustGroup)
```
foxcpp: Out of curiosity, is there a specific reason why this code is always iterating over all groups instead of just doing
```
if g.GetName() == a.mustGroup {
  return nil
}
```
?

internal/auth/netauth/netauth.go
the-maldridge: Right, and I think it should be because it is not clear in the tutorial when you should/should not log in as an unqualified local name.

internal/auth/netauth/netauth.go
the-maldridge: Yep.

internal/auth/netauth/netauth.go
the-maldridge: Yep.

internal/auth/netauth/netauth.go
the-maldridge: Mostly because many things in netauth are timing sensitive, and I'm in the habit of doing those kinds of compares.  I changed it per your suggestion to short-circuit on match.

the-maldridge:
@foxcpp anything else?

foxcpp:
GitHub is a mess. Merged manually into dev as [c102d62](https://github.com/foxcpp/maddy/commit/c102d6242b11cd62ede79bd1bead0acd02129acb).

foxcpp:
Thanks for your work!

PR #537, opened 2022-09-14
internal/storage/blob/s3: add iam, aws config file, minio config file configurable credentials

arush15june:
Implement configuration to choose the source of credentials for IMAP S3 blob storage. My primary purpose was to use the IAM profile instead of having plaintext AWS Access Keys. 

Along with IAM profile configuration, I have added configuration for the AWS config file and minio config file as well.


foxcpp:
Thanks!

PR #545, opened 2022-10-13
doc: replace `maddyctl` with `maddy`

pouriya:
There are 4 uses of `maddyctl` inside [dist/apparmor/dev.foxcpp.maddyctl](https://github.com/foxcpp/maddy/blob/master/dist/apparmor/dev.foxcpp.maddyctl) and 2 inside [internal/cli/app.go](https://github.com/foxcpp/maddy/blob/master/internal/cli/app.go#L99). I was not sure that I should remove/replace them too or not so I did not touch them.


foxcpp:
[dist/apparmor/dev.foxcpp.maddyctl](https://github.com/foxcpp/maddy/blob/master/dist/apparmor/dev.foxcpp.maddyctl) can be removed altogether. 

internal/cli/app.go should not be touched for at least one release - this is a backward-compatibility hack that makes maddy pretend to be maddyctl if it is symlinked as one.

pouriya:
Done

pouriya:
Should I remove [this line](https://github.com/pouriya/maddy/blob/master/build.sh#L142) or we need to create a link for at least one release?

pouriya:
ping

build.sh
```diff
@@ -139,7 +139,7 @@ install() {
 
 	command install -m 0755 -d "${destdir}/${prefix}/bin/"
 	command install -m 0755 "${builddir}/maddy" "${destdir}/${prefix}/bin/"
-	command ln -s maddy "${destdir}/${prefix}/bin/maddyctl"
+	command ln -s maddy "${destdir}/${prefix}/bin/maddy"
```
foxcpp: That's obviously broken

pouriya: Solved and rebased

PR #552, opened 2022-11-20
Doc: Update imapsql.md

felixkrohn:
From the actual behaviour I deduct that messages are stored under /data/messages, and this doc must be outdated?


foxcpp:
This is the Docker-specific location. For non-Docker installations, it is indeed /var/lib/maddy/messages.

Per https://maddy.email/docker/:
> Volumes
>
>/data - maddy state directory. Databases, queues, etc are stored here. You might want to mount a named volume there. The >main configuration file is stored here too (/data/maddy.conf).

foxcpp:
NACK

PR #561, opened 2023-01-02
fix: allow empty From address to match with TakeMsg

cfbraun:
With an empty from address TakeMsg is called, but ReleaseMsg is skipped.


foxcpp:
Thanks!

PR #566, opened 2023-01-28
Extract mapping lookup from loop

arisudesu:
Some repeated calls I've found while looking through prepare_email logic.


foxcpp:
Thanks!

Issue #369, opened 2021-07-08
Support HAProxy's PROXY protocol for SMTP and IMAP


foxcpp:
## Use case

Putting maddy behind a reverse proxy/load balancer/TLS terminator while still
providing necessary information for any policy validation.

## Implementation

See #296 for relevant discussion.

Exact configuration syntax and semantics is up for discussion.

## Task list

- [x] https://github.com/emersion/go-smtp/issues/147
- [x] https://github.com/emersion/go-smtp/pull/148
- [ ] Implement server support in endpoint/smtp
- [ ] Implement server support in endpoint/imap
- [ ] Implement client support in target.smtp (note that it would prevent pooling from working)

foxcpp:
Use-cases to evaluate:

## maddy behind a TLS terminator for all endpoints

TLS terminator needs to be protocol aware for server-server SMTP: It needs to proxy SMTP capabilities and inject STARTTLS support, implement it and pass through all necessary metadata (EHLO hostname). maddy needs to be aware of TLS being used for security policy evaluation.

For client-server SMTP and IMAP any special support in TLS terminator is not needed as long as implicit TLS is used which is recommended by IETF over STARTTLS anyway.

In addition, it might be useful to support TLS between proxy and maddy. We can probably limit this to implicit TLS to keep things simple.

## TCP load balancer with TLS handled by maddy

Simplest configuration to be supported.

foxcpp:
```
smtp tcp://127.0.0.1:2525 {
  proxy_protocol {
    assume_tls # should be turned on by default?
  }

  # if TLS is configured - it is for TLS between proxy and maddy. Not considered in policy evaluation.
  #tls ...

  ...
}

submission tcp://127.0.0.1:565 {
  proxy_protocol
  # insecure_auth assumed if proxy_protocol is used
}

imap tcp://127.0.0.1:1447 {
  proxy_protocol
  # insecure_auth assumed if proxy_protocol is used 
}
```

foxcpp:
Blocked by https://github.com/emersion/go-smtp/issues/147

infogulch:
This block seems to be resolved in https://github.com/emersion/go-smtp/pull/148

infogulch:
It looks like Caddy 2 currently supports the PROXY protocol via the [caddy-l4](https://github.com/mholt/caddy-l4j) app (see linked discussion). So Caddy is in good company with other proxies that will support this mail reverse-proxy configuration.

From OP:

[...]

PR #568, opened 2023-02-18
Proxy protocol support for SMTP and IMAP

drdaeman:
Hey, first of all, once again thank you for this project! I need to put maddy behind a reverse proxy (Traefik, but shouldn't really matter), but it needs to be aware about the original IP address, so I want it to support PROXY protocol.

What do you think about this implementation? I haven't fully battle-tested it yet, but it seem to work and to best of my awareness it should resolve two out of three remaining point of #369. Of course, I want this to be accepted, so I'm willing to make this PR as good as possible (if needed) and would appreciate any feedback.

Two potential issues I see are related to tests:
- First is that I've (ab)used SPF to test that the proxied IP address is recognized, as I'm not aware of any way to get it back. So PROXY tests assume that SPF is working correctly, which isn't exactly great. Your advice on this is most appreciated - maybe there is some debug command that I can issue to get test SMTP server to echo who it thinks is connecting?
- Another issue is that I cannot get tests to pass reliably. Even running one single test (`./run.sh -run TestProxyProtocolTrusted`) is flaky and half of the time results with a nondescript failure (all logs seem to be fine but it ends with "t.go:294: The server did not stop cleanly, deadlock?"). And when running multiple tests all together it seems that there's some interference between them and that trusted address list from one TestProxyProtocol function can somehow leak into another. I haven't investigated this further as it seems to be a different issue and could be just something about my local Go installation.

Thanks!


tests/smtp_test.go
```diff
@@ -68,6 +69,94 @@ func TestCheckRequireTLS(tt *testing.T) {
...
+		},
+	})
+	t.Port("smtp")
+	t.Config(`
+		smtp tcp://127.0.0.1:{env:TEST_PORT_smtp} {
+			hostname mx.maddy.test
+			tls off
+
+			proxy_protocol {
+				trust ` + tests.DefaultSourceIP.String() + ` ::1/128
```
drdaeman, line 87: IPv6 address here is to ensure parser recognizes both IPv4 (presuming `DefaultSourceIP` is IPv4) and IPv6.

tests/smtp_test.go
```diff
@@ -68,6 +69,94 @@ func TestCheckRequireTLS(tt *testing.T) {
...
+
+			deliver_to dummy
+		}
+	`)
+	t.Run(1)
+	defer t.Close()
+
+	conn := t.Conn("smtp")
+	defer conn.Close()
+	conn.Writeln(fmt.Sprintf("PROXY TCP4 192.168.99.1 %s 12345 %d", tests.DefaultSourceIP.String(), t.Port("smtp")))
```
drdaeman: Hardcoded IP address may be not exactly a best approach. Is this fine as it is, or should I add another constant (`SecondarySourceIP`)?

tests/smtp_test.go
```diff
@@ -68,6 +69,94 @@ func TestCheckRequireTLS(tt *testing.T) {
...
+	conn.SMTPNegotation("localhost", nil, nil)
+	conn.Writeln("MAIL FROM:<testing@one.maddy.test>")
+	conn.ExpectPattern("250 *")
```
drdaeman, line 111: As I've mentioned in the PR description, I'm abusing SPF checker to ensure that SMTPd thinks it serves a client from 192.168.99.1 and not 127.109.97.100. It surely works, although I'm not really happy about this and if someone can advise me a different approach I would be more than happy to try something else.

drdaeman:
I've battle-tested this on my own SMTP service running latest Maddy with my patches included for the last week, using `proxy_protocol` behind Traefik - and it seem to work without issues. All the headers and logs are spot on, no 192.168s. I haven't production-tested IMAP - I run Dovecot (because of dsync) - but I don't suppose it should be any different.

tests/smtp_test.go
foxcpp: This seems confusing.  Testing code binds on DefaultSourceIP before the connection is made and connects to 127.0.0.1.

Shouldn't it be `PROXY TCP4 <DefaultSourceIP> 127.0.0.1 12345 smtp`?

foxcpp: Probably using `127.0.0.17` would be better to avoid any traffic if test misbehaves and actually tries to contact 192.168.99.1.

docs/reference/endpoints/imap.md
```diff
@@ -36,6 +36,22 @@ tls cert.crt key.key {
...
 See [TLS configuration / Server](/reference/tls/#server-side) for details.
 
+**Syntax**: proxy_protocol _trusted ips..._ { ... } <br>
+**Default**: not enabled
+
+Enable use of HAProxy PROXY protocol. Supports both v1 and v2 protocols.
+If a list of trusted IP addresses or subnets is provided, only connections
+from those will be trusted.
+
+TLS for the channel between the proxies and maddy can be configured
```
foxcpp, line 50: Is `tls` top-level directive unused then?

drdaeman: Oh, good catch! Yes, that's correct, [`inheritGlobal` is `false`](https://github.com/drdaeman/maddy/blob/3910d91be11145312a418560edb72c710301f6e3/internal/endpoint/imap/imap.go#L87) for the `tls` directive in `ProxyProtocol`. In my own setup, I've used different set of certificates and haven't realized global TLS config is a thing. Unless you think it's fine the way it is, I will work on a change that would make global config respected (in a same manner `tls` directive works for IMAP).

foxcpp: In setupListeners in both SMTP and IMAP, created listener is wrapped using tls.NewListener using endpoint-level TLS config before proxy_protocol.NewListener is called. I believe proxy_protocol tls directive can be simply removed.

drdaeman:
Hey, sorry for a long time to respond - had a few things going on in my life that took all of my time. My bad.

I've rebased my branch and updated the tests, using the recommended 127.0.0.17 address - good consideration, thank you.

I gave it some thought and I'm a bit reluctant to remove the `tls` directive from `proxy_protocol`. Here's my logic:
- Normally, PROXY protocol is used between a load balancer (LB) and individual server (Maddy), and those are typically connected by a trusted network - so, usually one wouldn't use TLS on top of the LB-to-Maddy connection.

    LB ---PROXY[TLS[IMAP]]----> Maddy

  Here, global or IMAP (or SMTP)-level TLS would secure the communications inside the proxied connection, and the connection header would be sent as plaintext. Making `proxy_protocol`'s `tls` inherit the configuration from the top-level `tls` directive would result in odd behavior, where TLS is applied on top of a wrong layer.

- You are absolutely correct that if we'd just remove `tls` from `proxy_protocol` entirely, the issue would be moot.

- However, I believe there is a legit use case where having `tls` would be useful. In case the connection between the LB and Maddy is not trusted (e.g. LB is CloudFlare-like service and the connection is over Internet, not some private LAN/VPC), one may want to put TLS *on top* of PROXY protocol, so no one in position to observe or mess with the traffic between the LB and Maddy would be able to learn or do anything harmful:
  
    LB ---TLS[PROXY[IMAP]]----> Maddy
  
  or possibly even
    
    LB ---TLS_private[PROXY[TLS_public[IMAP]]]----> Maddy
  
  Which is why I'd suggest to keep the directive available, but required to be configured very explicitly, without any inheritance (I've amended docs slightly to be clear about this). It's a low-maintenance thing after all, because all the primitives (`tls2.TLSDirective` and `tls.NewListener`) needed to support it are already there.
  
  I admit, in such scenarios, one would most likely want to also have TLS client certificates for mutual authentication (so Maddy can authenticate the LB as well, possibly relying on PKI trust rather than IP whitelist), but I'd say that's some extra work that is kind of out-of-scope here for the base PROXY protocol support and that it could be done separately at some later point.

I hope this makes sense. If it doesn't, or you feel like that's not a scenario you want to support - let me know and I'll just nuke `tls` directive from there.

Again, sorry for a long silence and thanks!

foxcpp:
Thanks for your work!

PR #617, opened 2023-08-14
Update systemd services to depend on network-online.target

### LINKED ISSUE #616: Maddy systemd service tries to start before network interfaces are up
# Describe the bug

If maddy is managed by systemd using the systemd service file [in this repository](https://github.com/foxcpp/maddy/tree/master/dist/systemd), when the server reboots, maddy tries to start up too early, before network interfaces are ready. This leads to repeated failed startups, with the error message in logs:

```
Aug 13 14:42:33 mail2 maddy[1305]: smtp: listen tcp 111.111.111.111:25: bind: cannot assign requested address
Aug 13 14:42:33 mail2 systemd[1]: maddy.service: Main process exited, code=exited, status=1/FAILURE
Aug 13 14:42:33 mail2 systemd[1]: maddy.service: Failed with result 'exit-code'.
Aug 13 14:42:33 mail2 systemd[1]: Failed to start maddy mail server.
Aug 13 14:42:33 mail2 systemd[1]: maddy.service: Scheduled restart job, restart counter is at 4.
```

The systemd service has a `After=network.target` line, I think the fix is to use `After=network-online.target` instead.

From https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/:

> network.target has very little meaning during start-up. It only indicates that the network management stack is up after it has been reached. Whether any network interfaces are already configured when it is reached is undefined. [...]

> network-online.target is a target that actively waits until the nework is "up", where the definition of "up" is defined by the network management software. Usually it indicates a configured, routable IP address of some kind. Its primary purpose is to actively delay activation of services until the network is set up. [...]

# Steps to reproduce

Reboot the server, watch logs for maddy startup failures.

I tested with the `network-online` target too, and after server reboot maddy started on the first try.

# Configuration file

In maddy.conf, I have a line that sets a specific IP to bind to:

```
local_ip 111.111.111.111
```

cuu508:
Fixes: #616

### DIFF
--- a/dist/systemd/maddy.service
+++ b/dist/systemd/maddy.service
@@ -3,7 +3,7 @@ Description=maddy mail server
 Documentation=man:maddy(1)
 Documentation=man:maddy.conf(5)
 Documentation=https://maddy.email
-After=network.target
+After=network-online.target
 
 [Service]
 Type=notify
--- a/dist/systemd/maddy@.service
+++ b/dist/systemd/maddy@.service
@@ -3,7 +3,7 @@ Description=maddy mail server (using %i.conf)
 Documentation=man:maddy(1)
 Documentation=man:maddy.conf(5)
 Documentation=https://maddy.email
-After=network.target
+After=network-online.target
 
 [Service]
 Type=notify


foxcpp:
Thanks!

PR #622, opened 2023-08-22
fix(milter): remove erroneous path check

mmatous:
Hi, milter check doesn't currently accept paths to unix sockets. This should fix it. 
I've never written any Go before, so there may be some dumb oversight.


foxcpp:
Thanks!

PR #623, opened 2023-08-22
fix(f2b): use correct retry directive

mmatous:
`maxtries` is not a valid option for f2b jails, `maxretry` is. 
[proof](https://github.com/fail2ban/fail2ban/blob/e1d3006b0330e9777705a7baafe3989d442ed120/config/jail.conf#L107)


foxcpp:
Thanks!

PR #640, opened 2023-10-20
Correct spf dns record

opswill:
Based on the spf documnet: 
http://www.open-spf.org/action_browse_id_FAQ/Common_mistakes_revision_26/#helo
```
example.com.             IN  TXT  "v=spf1 mx -all"
mailserver.example.com.  IN  TXT  "v=spf1 a -all"
```
the second spf record in doc should set to a record. otherwise the mail-tester will show 
```
SPF_HELO_SOFTFAIL		SPF: HELO does not match SPF record (softfail)
```
softfail


foxcpp:
Thanks!

PR #642, opened 2023-10-27
fix: bump alpine to 3.18.4

onedr0p:
I am not sure if you want this to be `alpine:3.18` instead, feel free to change it.


foxcpp:
Thanks!

PR #643, opened 2023-10-30
docs/reference/modifiers/dkim: fix number of sigs for sign_fields

emersion:
oversign_fields is signed n+1 times, but sign_fields only n times.


foxcpp:
Thanks!

PR #647, opened 2023-11-11
Use distroless image instead of Alpine

ananthb:
Use distroless base image instead of an Alpine based image for the final layer.


foxcpp:
Do these images contain CA certificate bundles? They are necessary for proper mail server operation.

ananthb:
Yep, that's included. The [README](https://github.com/GoogleContainerTools/distroless/blob/main/base/README.md) for the base image mentions this, and I've personally used them before.

Issue #603, opened 2023-07-06
Empty "From" address results in rejection (5.1.7 Malformed address)


Dr-Escher:
# Describe the bug

While testing, I noticed maddy rejects mails with an empty "From" address.

First I wasn't sure about this behaviour, but RFC 1123 section 5.2.9 seems to demand the acceptance of such mails, as described here:
https://serverfault.com/questions/151955/why-an-empty-mail-from-address-can-sent-out-email/151987#151987

Also this was previously considered as an issue:
https://github.com/foxcpp/maddy/issues/337#issuecomment-783570586

# Steps to reproduce

In fact I triggered this using an Outlook auto responder, as mentioned in the answer linked above.

# Log files

Log from the sender:
```
from=<>, size=6781, class=0, nrcpts=1, msgid=<8b89858fca64442bb6508d83a24aae3b@FRYP281MB2793.DEUP281.PROD.OUTLOOK.COM>, proto=ESMTP, daemon=SMTP_MTA
dsn=5.1.7, stat=User unknown(Reason from remote:550 5.1.7 Malformed address (msg ID = 6249bab0))
```

Log from the receiver (maddy):
```
smtp: MAIL FROM error (deferred) {"check":"spf","msg_id":"d7d27c21","rcpt":"redacted@for.privacy","reason":"Malformed address","smtp_code":550,"smtp_enchcode":"5.1.7","smtp_msg":"Malformed address"}
```

# Configuration file

Default configuration except `$(hostname)` and `$(primary_domain)`

# Environment information

Running latest maddy from git master on Debian Bookworm 12.0

foxcpp:
Fixed in [ab94e0b](https://github.com/foxcpp/maddy/commit/ab94e0bb95428775f92bd0f25815595d1744f9c1).

PR #652, opened 2023-12-02
accept empty from-address in SPF fixes foxcpp/maddy#603

Beutlin:


foxcpp:
>[2.4](https://datatracker.ietf.org/doc/html/rfc7208#section-2.4).  The "MAIL FROM" Identity
>
>   SPF verifiers MUST check the "MAIL FROM" identity if a "HELO" check
   either has not been performed or has not reached a definitive policy
   result by applying the check_host() function to the "MAIL FROM"
   identity as the <sender>.
>
>   [RFC5321] allows the reverse-path to be null (see Section 4.5.5 in
   [RFC5321]).  In this case, there is no explicit sender mailbox, and
   such a message can be assumed to be a notification message from the
   mail system itself.  **When the reverse-path is null, this document
   defines the "MAIL FROM" identity to be the mailbox composed of the
   local-part "postmaster" and the "HELO" identity (which might or might
   not have been checked separately before).**

RFC 7208 has this requirement for this case, in case of empty MAIL FROM it seems like it should be assumed to be equal to "postmaster@HELO-hostname-here". That is, "postmaster@" + s.msgMeta.Conn.Hostname.

foxcpp:
Superseded by [ab94e0b](https://github.com/foxcpp/maddy/commit/ab94e0bb95428775f92bd0f25815595d1744f9c1).

PR #662, opened 2024-01-09
Fix escape on authorization page

rebane2001:
The asterisks on the page are not escape, leading to "" instead of "\*" when rendered.


foxcpp:
Thanks!

PR #663, opened 2024-01-10
Bump go-smtp to fix SMTP smuggling

### LINKED ISSUE #661: Strict CRLF check in SMTP protocol
The `go-smtp` package has [released a new version](https://github.com/emersion/go-smtp/releases/tag/v0.20.0), where [strict CRLF check](https://github.com/emersion/go-smtp/issues/243) is performed to mitigate an zero-day attack against SMTP known as [SMTP Smuggling](https://www.postfix.org/smtp-smuggling.html). I think Maddy should probably update the dependency and (probably) release a new version with the updated dependency.

* maddy version: all?

McSinyx:
This should resolve GH-661.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/emersion/go-milter v0.3.3
 	github.com/emersion/go-msgauth v0.6.6
 	github.com/emersion/go-sasl v0.0.0-20220912192320-0145f2c60ead
-	github.com/emersion/go-smtp v0.16.0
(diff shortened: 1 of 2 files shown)


foxcpp:
There are some API changes in go-smtp, need to change method signatures to match them.

McSinyx:
Thanks for point that out.  I can't get the test suit succeed locally for the base commit (nor for v0.7.0) and kinda bet on the bump to work.  Could you please take this over?

PR #667, opened 2024-01-18
Update remote.md

fluidum:
1. fixed security level links (white pages)
2. changed default of min_tls_level (https://github.com/foxcpp/maddy/blob/master/maddy.conf)


foxcpp:
Thanks!

PR #670, opened 2024-01-25
Removing duplicated test case

ToutPetitAdrien:
Test cases 15 and 16 are duplicates. I'm removing one of them here 🧹


foxcpp:
Thanks!

Issue #619, opened 2023-08-17
Gandi dns-01 challenge fail: 400 Absolute rrset_name must end with mydomain.org


onny:
# Describe the bug

Getting following error trying to renew tls cert with dns-01 challenge and Gandi: `could not get certificate from issuer        {"error":"[example.org] solving challenges: presenting for challenge: adding temporary record for zone \"example.org.\": LiveDNS returned a 400 (Absolute rrset_name must end with example.org. (_acme-challenge.)`

Was already using it for a while like this but renewing the cert seems to fail

# Log files

```
$(hostname) = mx1.example.org
$(primary_domain) = example.org
$(local_domains) = $(primary_domain)
hostname mx1.example.org

tls {
  loader acme {
    email onny@project-insanity.org
agreed
hostname mx1.example.org
challenge dns-01
dns gandi {
  api_token "{env:GANDI_API_KEY}"
}

  }
}
```

# Configuration file

```
Aug 17 15:07:56 piproxy maddy[759807]: tls.loader.acme/acme_client: trying to solve challenge        {"ca":"https://acme-v02.api.letsencrypt.org/directory","challenge_type":"dns-01","identifier":"example.org"}
Aug 17 15:07:57 piproxy maddy[759807]: tls.loader.acme/acme_client: cleaning up solver        {"challenge_type":"dns-01","error":"no memory of presenting a DNS record for \"_acme-challenge.example.org\" (usually OK if presenting also failed)","identifier":"example.org"}
Aug 17 15:07:57 piproxy maddy[759807]: tls.loader.acme/obtain: could not get certificate from issuer        {"error":"[example.org] solving challenges: presenting for challenge: adding temporary record for zone \"example.org.\": LiveDNS returned a 400 (Absolute rrset_name must end with example.org. (_acme-challenge.)) (order=https://acme-v02.api.letsencrypt.org/acme/order/1114545927/202211442616) (ca=https://acme-v02.api.letsencrypt.org/directory)","identifier":"example.org","issuer":"acme-v02.api.letsencrypt.org-directory"}
```

# Environment information

* maddy version: 0.7

nustiueudinastea:
Hitting the same issue

foxcpp:
[d9920f0](https://github.com/foxcpp/maddy/commit/d9920f0763cdaf6e49d015732b0de499299fbace) might fix it, but not sure tbh.

[...]

Issue #555, opened 2022-11-23
verifying github.com/libdns/gandi@v1.0.2: checksum mismatch


sewnie:
# Describe the bug

can't build maddy

# Steps to reproduce
```
git clone https://github.com/foxcpp/maddy.git
cd maddy
./build.sh
```

# Log files

```
-- Version: 0.6.2+e1b3391
-- Building main server executable...
go: downloading github.com/caddyserver/certmagic v0.16.1
go: downloading github.com/urfave/cli/v2 v2.10.2
go: downloading github.com/foxcpp/go-imap v1.0.0-beta.1.0.20220623182312-df940c324887
go: downloading github.com/foxcpp/go-imap-sql v0.5.1-0.20220627220518-df3b66a5b04f
go: downloading golang.org/x/crypto v0.0.0-20220622213112-05595931fe9d
go: downloading github.com/klauspost/cpuid/v2 v2.0.14
go: downloading github.com/libdns/libdns v0.2.1
go: downloading github.com/mholt/acmez v1.0.2
go: downloading github.com/miekg/dns v1.1.50
go: downloading go.uber.org/zap v1.21.0
go: downloading golang.org/x/net v0.0.0-20220622184535-263ec571b305
go: downloading github.com/emersion/go-message v0.16.0
go: downloading github.com/emersion/go-msgauth v0.6.6
go: downloading github.com/emersion/go-smtp v0.15.1-0.20220119142625-1c322d2783aa
go: downloading github.com/emersion/go-sasl v0.0.0-20211008083017-0b9dcfb154ac
go: downloading github.com/foxcpp/go-dovecot-sasl v0.0.0-20200522223722-c4699d7a24bf
go: downloading github.com/go-ldap/ldap/v3 v3.4.3
go: downloading golang.org/x/text v0.3.7
go: downloading github.com/GehirnInc/crypt v0.0.0-20200316065508-bb7000b8a962
go: downloading golang.org/x/sync v0.0.0-20220601150217-0de741cfad7f
go: downloading github.com/emersion/go-milter v0.3.3
go: downloading blitiri.com.ar/go/spf v1.5.1
go: downloading github.com/emersion/go-imap-compress v0.0.0-20201103190257-14809af1d1b9
go: downloading github.com/emersion/go-imap-sortthread v1.2.0
go: downloading github.com/foxcpp/go-imap-i18nlevel v0.0.0-20200208001533-d6ec88553005
[...]

PR #673, opened 2024-01-26
upgrade Gandi libdns to 1.0.3

nustiueudinastea:
upgraded to the latest version of gandi libdns, which fixes an issue where new records could not be created

See https://github.com/foxcpp/maddy/issues/619


foxcpp (changes requested):
Looks like there are some incompatible changes in latest commits - https://github.com/libdns/gandi/commit/5a260222d1a95412b414346cbccec30c885a55cc
I guess it is a mess we will have to bring downstream by breaking config compatibility as it is no longer possible to specify an API key in libdns/gandi despite them [being still supported](https://api.gandi.net/docs/authentication/).

Also, libdns/gandi maintainer overwrites existing releases, leading to issues with Go tooling (#555, libdns/gandi#9), therefore it is important to specify a particular commit in go.mod using pseudoversions.

foxcpp:
Thanks for your work!

PR #682, opened 2024-02-15
(maybe) Fixes a typo in the `rspamd` docs

reind33r:
Currently, to add _rspamd_, [the docs ](https://maddy.email/third-party/rspamd/) suggest to

> add rspamd check with default configuration into appropriate check block (probably in local_routing):
> ```
> checks {
>    ...
>    rspamd
> }```
>

Adding this block in `local_routing` results in a failure to start, because in this context only `check` is valid.

(I see that `checks` exist at top-level, but if we mention `local_routing` and appropriate `check` block, it makes more sense to me that the code block is also `check`)

### DIFF
--- a/docs/third-party/rspamd.md
+++ b/docs/third-party/rspamd.md
@@ -7,7 +7,7 @@ If rspamd is running locally, it is enough to just add `rspamd` check
 with default configuration into appropriate check block (probably in
 local_routing):
 ```
-checks {
+check {
     ...
     rspamd
 }


foxcpp:
It is indeed a typo.

`checks` and `modifiers` are names of group modules that allow defining reusable configurations for `check` and `modify` directives correspondingly:

```
checks inbound_filters {
        dkim
}

modifiers inbound_mutation {
        replace_rcpt regexp "asd" "asd"
}

smtp tcp://127.0.0.1:2525 {
        check &inbound_filters
        modify &inbound_mutation
        reject
}
```

foxcpp:
Thanks!

PR #686, opened 2024-03-13
build: make "./build.sh install" reusable

oidq:
* prevent the script from blindly overwriting the current configuration
* force maddyctl symbolic linking
* fix indentation in build.sh

Few days ago I learned the hard way, that running `./build.sh install` on existing installation of maddy blindly overwrites the configuration. This commit alters the bahevior of `build.sh` to not blindly copy over existing `maddy.conf` and force the creation of the `maddyctl` symlink so that the script can be used to update current installation.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -3,6 +3,7 @@
 destdir=/
 builddir="$PWD/build"
 prefix=/usr/local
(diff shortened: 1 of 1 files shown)


foxcpp:
Thanks!

PR #689, opened 2024-03-16
Add ACME-DNS provider for libdns

rjocoleman:
This PR adds the ACME-DNS provider for libdns.

References:

* https://github.com/libdns/acmedns
* https://github.com/joohoi/acme-dns?tab=readme-ov-file#acme-dns

In short ACME-DNS is a project that allows delegation of challenges to an acme-dns server which can be used in the case of unsupported provider  or for particular security models.

This commit also fixes a bug in the docs, I can pull it out to a seperate commit if preferred: https://github.com/foxcpp/maddy/pull/689/files#diff-72bbc317ad42981064e598737075d5acf9e7b0d88c58d1e46f3ff09e0486bb45R152


foxcpp:
Provider looks lightweight enough so we can probably enable it by default (will adjust your PR to do that). Also to promote ACME-DNS use, it looks like a good thing)

PR #690, opened 2024-03-16
Add additional build tags to Dockerfile

rjocoleman:
This PR adds support for passing additional tags to the `./build.sh` inside the `Dockerfile` at build time.

Use case, as per the docs I've added, is to be able to get more libdns providers in there (like #689)


foxcpp:
Thanks!

PR #708, opened 2024-06-03
fix possible missing "not" from documentation

w1kee:
this changes the meaning of the sentence, but i think the "not" was intended to be there because the sentence sounds weird without it


foxcpp:
Thanks!

PR #716, opened 2024-07-14
Fix misspelling in setting-up.md

naltun:
I'm installing Maddy on OpenBSD and I'm having fun with it. I'll report back if I find anything else.


foxcpp:
Actually, that was a little joke. :)  

^W is how a terminal would print Ctrl-W as a character (you must have seen it with Ctrl-C being ^C at least once). 
Ctrl-W removes the last word.  So it is like "use ~~vim~~ your favorite editor" but more elaborate.

naltun:
Programming humor, nicely done :handshake: 

Nice project :)

Issue #684, opened 2024-02-25
Documentation for aliases


dehnhard:
# Describe the bug

Missing documentation of `/etc/maddy/aliases`.

# Steps to reproduce

I followed the tutorial to setup a simple personal mail server. It was easy and straightforward. Just the needed mail aliases took a long time to setup, because it was not documented and a search brought me at first to some complex setup descriptions. Eventually I found a mention of the aliases file in some PR or commit message.

# Log files
n/a

# Configuration file
n/a

# Environment information
* maddy version: 0.7.1

A sentence about `/etc/maddy/aliases` at the end of the tutorial and/or an example file for e.g. an alias hostmaster would be nice for newcomers.

stellarpower:
It would be helpful to have a quick snippet of what the file should look like in the docs page. I had this working at some point but that file got lost, and now trying a few obvious permutations to recreate it and all don't seem to be working for some reason. Thanks

PR #729, opened 2024-10-14
Point towards syntax of (aliases) file table for new users.

stellarpower:
#684


stellarpower:
A reference may want to be added in other locations too, I figured the example config file was probably going to be spotted by anyone who had questions on how to use it. 

I also had mine without the colon, i.e. `alias@localhost real-recipient@localhost`, and this was parsed fine as the mapping `"alias@localhost real-recipient@localhost" => ""`, which should perhaps be a parse error for the context of address rewriting and throw an error, or at least a warning.

Issue #736, opened 2024-12-09
DMARC align check may broken if TLD is used to send mails


mschneider82:
I had some dkim alignment issues with mail domains listed as a TLD in https://publicsuffix.org/list/public_suffix_list.dat
For example: if gitlab.io is used to send DKIM DMARC mails, with a policy adkim=r restricted policy, the isAligned() function will return an error because it cannot find a domain:

https://github.com/foxcpp/maddy/blob/master/internal/dmarc/evaluate.go#L212

It can be fixed by adding in line 209:

```
	tld, _ := publicsuffix.PublicSuffix(fromDomain)
	if strings.EqualFold(fromDomain, tld) {
		return strings.EqualFold(fromDomain, authDomain)
	}
```

If you want, i can make a PR for that, or if you have another solution feel free! thanks!
I know the DKIM RFC doesnt handle this case well and require to use a Organisational Domain but there are quite a few big goverment domains used for mailing in that list.

mschneider82:
fixed in #737

PR #737, opened 2024-12-09
#736: dmarc evalute added check if fromDomain is a TLD

mschneider82:
fixes the use of a TLD with DMARC DKIM policy relaxed


foxcpp:
Thanks!

PR #758, opened 2025-02-08
Build aarch64 binary in release workflow

### LINKED ISSUE #752: Release Prebuilt AArch64 Binaries
# Use Case

Maddy currently does not provide prebuilt binaries for the aarch64 architecture. This limits the ability to run Maddy on popular ARM-based systems.

# Alternatives Considered

**Building from source**: While possible, this adds complexity, especially for users unfamiliar with Go build toolchains. I am personally building from source at the moment, but this adds extra maintenance overhead.

# Your Idea for a Solution

Maddy should provide official aarch64 binaries as part of its release pipeline.

## Implementation Details

- **CI/CD Pipeline Enhancement**: With the general availability of ARM64 runners on GitHub Actions, the existing CI/CD pipeline can be extended to natively build for aarch64.

- **Testing**: Utilize GitHub's ARM64 runners to test the binaries, ensuring stability and performance on ARM-based systems.

- **Distribution**: Distribute the binaries in the same manner as x86_64 builds.

[ &#9989;] I'm willing to help with the implementation

nurmukhametov:
Add an extra job to the release pipeline to build the aarch64 binary.

This PR fixes #752.

Building in a Docker container is a workaround for the issue of JavaScript-based GitHub Actions not being supported in Alpine containers on the Arm64 platform. Otherwise, we could completely reuse artifact-builder-x86 as a matrix job by running it on an Arm runner. This could be done later when upload-artifact works in an Alpine Arm64 container.

This issue https://github.com/actions/runner/issues/801 in Action runners seems to be related, and its potential fix is here (https://github.com/actions/runner/pull/3665).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,8 +11,8 @@ permissions:
   packages: write
 
 jobs:
-  artifact-builder:
-    name: "Prepare release artifacts"
+  artifact-builder-x86:
+    name: "Prepare release artifacts (x86)"
     if: github.ref_type == 'tag'
     runs-on: ubuntu-latest
     container:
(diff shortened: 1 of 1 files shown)


foxcpp:
Thanks! Will merge and see how it works with the next release.

PR #767, opened 2025-02-22
target/remote: Add smtp_ports option for remote targets

johnweldon:
Allow ports other than 25 to be used for outgoing SMTP attempts.

This change keeps the default behavior the same, but allows overriding it with configuration.

(Will rebase if/after #766 is merged)

### DIFF
--- a/dist/vim/syntax/maddy-conf.vim
+++ b/dist/vim/syntax/maddy-conf.vim
@@ -198,6 +198,7 @@ syn keyword maddyModDir
 	\ sig_expiry
 	\ sign_fields
 	\ sign_subdomains
+	\ smtp_ports
 	\ softfail_action
 	\ SOME_action
 	\ source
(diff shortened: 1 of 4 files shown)


foxcpp:
NACK. Ports 587 and 465 are used for mail *submission* (as defined by RFC 4409) by user agents, not for mail *relay* that uses exclusively port 25. While workarounds such as this PR may work in some situations, this is non-standard behavior. Besides, recommending to use 587 over default 25 brings no additional security. The correct course of action is to request port 25 unblock from ISP or to forward outbound messages via an external host (usually ISP-provided). In my experience, many VPS providers are willing to unblock port 25 when requested via support ticket. Notable exceptions include "cloud" providers but most of them also provide outbound email services that can serve a as a proxy.

johnweldon:
Is there any benefit to this feature if it's documented differently, or would you rather not merge it in any form?

Thanks for the review.

foxcpp:
I don't think there is any benefit in this feature.

Issue #650, opened 2023-11-27
POP3 support


kiwiz:
# Use case

Interfacing with clients that only support POP3 (Gmail, for example)

# Your idea for a solution

Adding POP3 support to maddy

- [x] I'm willing to help with the implementation

foxcpp:
I would simply map POP3 operations to corresponding IMAP ones and reuse same storage interfaces and so on. We need to have a reasonably robust library to decode/encode wire protocol, the rest should be fairly easy.

PRs are welcome, let me know if you need any guidance on where to start.

foxcpp:
https://github.com/regnull/popgun might be a good starting point.

kiwiz:
A couple of questions/comments:

* Implementing this is as easy* as creating `internal/endpoint/pop3/pop3.go` and importing the package in `maddy.go`?
* Is there a preference for integrating the logic into maddy vs just depending on `regnull/popgun`?
    * My guess is the former as `popgun` hardcodes the default logger and doesn't let you specify multiple listeners

foxcpp:
>Implementing this is as easy* as creating internal/endpoint/pop3/pop3.go and importing the package in maddy.go

Would need to figure how to map IMAP-specific storage interface to POP3 operations. Might be:
* Lock -> Select mailbox
* Unlock -> Close mailbox
* Update -> Expunge
* Rset -> ???
* Dele -> Store +FLAGS \\Deleted
* List, Retr, Top -> Fetch

Lock/Unlock would also need to create per-user lock somehow, not exactly sure how that would work.


>Is there a preference for integrating the logic into maddy vs just depending on regnull/popgun?
>>My guess is the former as popgun hardcodes the default logger and doesn't let you specify multiple listeners

I would fork popgun and do necessary adjustments to support pluggable logger and multiple listeners.  And then use that as a library in main codebase.

ehfd:
This would be beneficial for using custom email addresses within Gmail, since IMAP is only supported with Gmailify (Yahoo, AOL, etc.).

lukastribus:
Unfortunately Gmail is abandoning both Gmailify and POP import: https://support.google.com/mail/answer/16604719?hl=en

> Fetching emails from third-party accounts into your Gmail account, with POP, will no longer be supported.

foxcpp:
I am working currently on go-imap v2 migration. This will take a while but I intend to merge #769 along the way while adjusting it match new IMAP interfaces.

PR #769, opened 2025-03-02
WIP: POP3 support

kiwiz:
Adds support for the POP3 protocol. A lot of work is still needed, but I wanted to get some early feedback on the direction I'm taking. The code was based off `imap`, so there may be some dead code remaining.

Implements #650

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -107,6 +107,7 @@ require (
 	github.com/hashicorp/hcl v1.0.0 // indirect
 	github.com/jmespath/go-jmespath v0.4.0 // indirect
 	github.com/josharian/intern v1.0.0 // indirect
+	github.com/kiwiz/popgun v0.0.0 // indirect
 	github.com/klauspost/compress v1.17.11 // indirect
 	github.com/klauspost/cpuid/v2 v2.2.9 // indirect
 	github.com/magiconair/properties v1.8.9 // indirect
@@ -171,3 +172,5 @@ replace github.com/emersion/go-imap => github.com/foxcpp/go-imap v1.0.0-beta.1.0
 replace github.com/emersion/go-smtp => github.com/foxcpp/go-smtp v1.21.4-0.20250124171104-c8519ae4fb23 // v1.21.3+maddy.1
 
 replace github.com/libdns/gandi => github.com/foxcpp/libdns-gandi v1.0.4-0.20240127130558-4782f9d5ce3e // v1.0.3+maddy.1
+
+replace github.com/kiwiz/popgun v0.0.0 => ../popgun
--- a/internal/endpoint/pop3/pop3.go
+++ b/internal/endpoint/pop3/pop3.go
@@ -444,23 +444,6 @@ func (endp *Endpoint) Unlock(user pop3backend.User) error {
 	return backendUser.Logout()
 }
 
-func (endp *Endpoint) I18NLevel() int {
-	be, ok := endp.Store.(i18nlevel.Backend)
-	if !ok {
-		return 0
-	}
-	return be.I18NLevel()
-}
-
-func (endp *Endpoint) SupportedThreadAlgorithms() []sortthread.ThreadAlgorithm {
-	be, ok := endp.Store.(sortthread.ThreadBackend)
-	if !ok {
-		return nil
-	}
-
-	return be.SupportedThreadAlgorithms()
-}
-
 func init() {
 	module.RegisterEndpoint("pop3", New)
 }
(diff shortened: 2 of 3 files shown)


internal/endpoint/pop3/pop3.go
```diff
@@ -0,0 +1,466 @@
...
+	modconfig.Table(cfg, "auth_map", true, false, nil, &endp.saslAuth.AuthMap)
+	if _, err := cfg.Process(); err != nil {
+		return err
+	}
+
+	if updBe, ok := endp.Store.(updatepipe.Backend); ok {
+		if err := updBe.EnableUpdatePipe(updatepipe.ModeReplicate); err != nil {
+			endp.Log.Error("failed to initialize updates pipe", err)
+		}
+	}
```
kiwiz: The `updatepipe` stuff is entirely unused. I'm not sure if there's any important side effects or if it can be removed.

internal/endpoint/pop3/pop3.go
```diff
@@ -0,0 +1,466 @@
...
+
+	seqset := imap.SeqSet{}
+	seqset.AddNum(uint32(msgId))
+	return mailbox.UpdateMessagesFlags(true, &seqset, imap.SetFlags, false, []string{imap.DeletedFlag})
+}
+
+// Undelete all messages marked as deleted in single connection
+func (endp *Endpoint) Rset(user pop3backend.User) error {
+	return fmt.Errorf("pop3: unimplemented")
```
kiwiz: Not sure how this can be implemented. Perhaps keep track of all messages that were Deleted and revert if this is called?

internal/endpoint/pop3/pop3.go
```diff
@@ -0,0 +1,466 @@
...
+// POP3 server sends the headers of the message, the blank
+// line separating the headers from the body, and then the
+// number of lines of the indicated message's body, being
+// careful to byte-stuff the termination character (as with
+// all multi-line responses).
+// Note that if the number of lines requested by the POP3
+// client is greater than than the number of lines in the
+// body, then the POP3 server sends the entire message.
+func (endp *Endpoint) Top(user pop3backend.User, msgId int, n int) (lines []string, err error) {
+	return nil, fmt.Errorf("pop3: unimplemented")
```
kiwiz: Is there code I can reference for formatting the message headers?

internal/endpoint/pop3/pop3.go
```diff
@@ -0,0 +1,466 @@
...
+}
+
+func (endp *Endpoint) SupportedThreadAlgorithms() []sortthread.ThreadAlgorithm {
+	be, ok := endp.Store.(sortthread.ThreadBackend)
+	if !ok {
+		return nil
+	}
+
+	return be.SupportedThreadAlgorithms()
+}
```
kiwiz: Not sure if these methods are necessary.

kiwiz:
The server currently compiles and runs, but I haven't tested it with a non-empty mailbox. Is there an easy way to generate/import test data for that purpose?

internal/endpoint/pop3/pop3.go
```diff
@@ -0,0 +1,466 @@
...
+	if err != nil && err != mess.ErrNoMessages {
+		return "", err
+	}
+
+	msg, ok := <- c
+	if !ok {
+		return "", fmt.Errorf("internal server error")
+	}
+
+	return strconv.FormatUint(uint64(msg.Uid), 10), nil
```
kiwiz: Return value is wrong.

internal/endpoint/pop3/pop3.go
foxcpp: These are relevant for IMAP implementation only and can be safely removed.

internal/endpoint/pop3/pop3.go
foxcpp: It is used for IMAP update synchronization when multiple servers are using the same DB, can be safely removed for POP3.

internal/endpoint/pop3/pop3.go
foxcpp: That seems to be most reliable solution, considering current IMAP-based interface.

internal/endpoint/pop3/pop3.go
foxcpp: Hmm, that one is tricky. I guess, you would need to consume Literals returned by ListMessages and format them yourself. It is okay to partially consume returned value.

go.mod
```diff
@@ -171,3 +172,5 @@ replace github.com/emersion/go-imap => github.com/foxcpp/go-imap v1.0.0-beta.1.0
 replace github.com/emersion/go-smtp => github.com/foxcpp/go-smtp v1.21.4-0.20250124171104-c8519ae4fb23 // v1.21.3+maddy.1
 
 replace github.com/libdns/gandi => github.com/foxcpp/libdns-gandi v1.0.4-0.20240127130558-4782f9d5ce3e // v1.0.3+maddy.1
+
+replace github.com/kiwiz/popgun v0.0.0 => ../popgun
```
kiwiz, line 176: These references will need to be updated

ehfd:
Any progress? It would be really nice to have this.

PR #771, opened 2025-03-02
feat(modify): new modifier `add_header`

### LINKED ISSUE #158: Add modifiers to rewrite headers
See https://github.com/foxcpp/maddy/issues/158#issuecomment-551166366.

--
Original post:

This should not be used for relays (RFC 5321 prohibits modification of existing header fields) but is usually fine for submission servers. In case of submission, this is what people probably want when saying "replace sender" \*.
Need to think whether it may be useful for replace_rcpt too.

```
replace_sender this that {
  header on # probably default
}
```

\* replacing only MAIL FROM command argument (as currently done by replace_sender) can invalidate DMARC alignment, btw.

msessa:
Partially addresses #158 

## Description

This PR adds a new modifier `add_header` which allows arbitrary headers to be added to a message in the pipeline.

## Use Case

A practical use case is to inject the `X-SES-FROM-ARN` header before forwarding the message to AWS SES to control which identity will be used to verify the sender domain.

See reference [here](https://docs.aws.amazon.com/ses/latest/dg/sending-authorization-delegate-sender-tasks-email.html)

### DIFF
--- a/.mkdocs.yml
+++ b/.mkdocs.yml
@@ -53,6 +53,7 @@ nav:
       - SMTP modifiers:
           - reference/modifiers/dkim.md
           - reference/modifiers/envelope.md
+          - reference/modifiers/header.md
       - Lookup tables (string translation):
           - reference/table/static.md
           - reference/table/regexp.md
--- a/docs/reference/modifiers/header.md
+++ b/docs/reference/modifiers/header.md
@@ -1,0 +1,23 @@
+# Header Modifiers
+
+## Adding a new header
+
+`add_header` module modifies the message by adding an header.
+
+Note: the header must be present in the message prior to the execution of the modifier. 
+If the header already exist it will result in an error.
+
+
+Definition:
+
+```
+add_header <headerName> <headerValue>
+```
+
+Use examples:
+
+```
+modify {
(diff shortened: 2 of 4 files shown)


foxcpp (changes requested):
Overall LGTM. Can you add a configuration option to not return an error on duplicate header? It is correct to have multiple values for some header fields and such configuration option may be useful for some cases. I would probably make it default even as "least surprise" principle.

msessa:
@foxcpp thanks for you review (and for creating maddy!)

I didn't realize the spec permitted fields appearing multiple times.
I've adapted the code and tests for that

msessa:
btw I'm happy to also implement other variations of the modifier, eg: `upsert_header` `remove_header` if you think they could be useful

msessa:
@foxcpp any update on this PR?

DEVisions:
Interested in this feature as well. Hope it gets merged :)

msessa:
@foxcpp 
Would it be possible to merge this?

foxcpp:
Needs rebase on latest dev, there are some module interface changes. Other than - sure, we can merge this. Sorry for the long wait I wasn't sure whether it is a good idea to it as-is.

PR #780, opened 2025-04-15
Implement maddy_queue_length metric

spiarh:
The maddy_queue_length was already registered but never used.

The gauge is updated when actions on disks are performed:

- Incremented when a message is written to disk or when the queue
is read from disk.
- Decremented when a message is removed from disk.


spiarh:
@foxcpp could you please take a look? that should be pretty quick to review ;)

spiarh:
@foxcpp sorry for the late reply, the tests may be a bit flaky. I did run the tests successfully on master and on this branch.

PR #802, opened 2025-10-25
Fix typo in TLS certificates section doc

d3vw:


foxcpp:
Fine, let's fix it) Thanks

PR #813, opened 2026-01-10
feat(dnsbl): Add per-response-code scoring with custom messages

thisisjaymehta:
## Summary

This PR adds support for per-response-code scoring in DNSBL checks, allowing administrators to assign different scores and custom rejection messages based on specific DNSBL return codes.  This enables efficient use of combined DNSBLs like Spamhaus ZEN with granular control. 

## Problem

Combined DNSBLs like Spamhaus ZEN (`zen.spamhaus.org`) return different response codes for different listing types: 

| Response Code | List | Severity |
|---------------|------|----------|
| `127.0.0.2`, `127.0.0.3` | SBL (known spam sources) | High |
| `127.0.0.4`-`127.0.0.7` | XBL (exploited/compromised hosts) | High |
| `127.0.0.10`, `127.0.0.11` | PBL (dynamic IPs) | Lower |

Currently, Maddy: 
1. Counts any response within `127.0.0.1/24` as a single "hit"
2. Applies the same score regardless of which specific code was returned
3. Cannot provide response-code-specific rejection messages
4. If multiple codes are returned, they all count as one hit with one score

Users who want different scores for different listing types must query separate lists (`sbl.spamhaus.org`, `xbl.spamhaus.org`, `pbl.spamhaus.org`), resulting in **3 DNS queries instead of 1**.

Reference: https://docs.spamhaus.com/datasets/docs/source/40-real-world-usage/PublicMirrors/MTAs/020-Postfix. html

## Solution

Add a new `response` configuration block that allows per-response-code scoring and custom messages:

```
check.dnsbl {
    reject_threshold 10
    quarantine_threshold 5

    zen.spamhaus.org {
        client_ipv4 yes
        client_ipv6 yes
        
        # SBL - Spamhaus Block List (known spam sources)
        response 127.0.0.2 127.0.0.3 {
            score 10
            message "Listed in Spamhaus SBL. See https://check.spamhaus.org/"
        }
        
        # XBL - Exploits Block List (compromised hosts)
        response 127.0.0.4 127.0.0.5 127.0.0.6 127.0.0.7 {
            score 10
            message "Listed in Spamhaus XBL. See https://check.spamhaus.org/"
        }
        
        # PBL - Policy Block List (dynamic IPs)
        response 127.0.0.10 127.0.0.11 {
            score 5
            message "Listed in Spamhaus PBL. See https://check.spamhaus.org/"
        }
    }
}
```

When multiple response codes are returned (e.g., `127.0.0.2` and `127.0.0.11`), their scores are **summed** (10 + 5 = 15 in this example).

### Data Structures
- Added `ResponseRule` struct with `Networks`, `Score`, and `Message` fields
- Extended `List` struct with `ResponseRules []ResponseRule`
- Extended `ListedErr` with `Score` and `Message` fields

### Core Logic
- `checkIP()` now matches response codes against rules and sums scores
- `checkLists()` uses `ListedErr.Score` when set, falls back to legacy `ScoreAdj`
- `ListedErr.Fields()` uses custom message when available

### Configuration
- Added `parseResponseRule()` to parse new `response` blocks
- Updated `readListCfg()` to handle response blocks via `AllowUnknown()`

### Backwards Compatibility
- Legacy `responses` + `score` configuration works unchanged
- New behavior only activates when `ResponseRules` is configured

## Benefits

| Feature | Before | After |
|---------|--------|-------|
| DNS queries for ZEN | 1 (codes ignored) | 1 (codes interpreted) |
| Per-code scoring | ❌ | ✅ |
| Custom rejection messages | ❌ | ✅ |
| Backwards compatible | N/A | ✅ |

## Testing

- Added test cases for multiple return codes with different scores
- Added test cases for low-severity vs high-severity responses
- Verified legacy configuration behavior is preserved

### DIFF
--- a/docs/reference/checks/dnsbl.md
+++ b/docs/reference/checks/dnsbl.md
@@ -216,7 +216,9 @@ and contains the following directives:
 **Required**
 
 Score to add when this response code is returned. If multiple response codes
-are returned by the DNSBL, scores are summed together.
+are returned by the DNSBL, and they match different rules, the scores from
+all matched rules are summed together. Each rule is counted only once, even
+if multiple returned IPs match networks within that rule.
 
 #### message _string_
 **Optional**
@@ -244,6 +246,12 @@ zen.spamhaus.org {
     }
 }
 ```
+
+**Scoring behavior:**
+- If DNSBL returns `127.0.0.2` only → Score: 10 (matches first rule)
+- If DNSBL returns `127.0.0.11` only → Score: 5 (matches second rule)
+- If DNSBL returns both `127.0.0.2` and `127.0.0.11` → Score: 15 (both rules match, scores sum)
+- If DNSBL returns both `127.0.0.2` and `127.0.0.3` → Score: 10 (same rule matches, counted once)
 
 **Backwards compatibility:** When `response` blocks are not used, the legacy
 `responses` and `score` directives work as before.
--- a/docs/reference/checks/dnsbl.md
+++ b/docs/reference/checks/dnsbl.md
@@ -207,6 +207,9 @@ Defines per-response-code rules for scoring and custom messages. This is useful
 for combined DNSBLs like Spamhaus ZEN that return different codes for different
 listing types.
 
+This works for both IP-based lookups (client_ipv4, client_ipv6) and domain-based
+lookups (ehlo, mailfrom).
+
 Each `response` block takes one or more IP addresses or CIDR ranges as arguments
 and contains the following directives:
 
--- a/docs/reference/checks/dnsbl.md
+++ b/docs/reference/checks/dnsbl.md
@@ -203,8 +203,6 @@ rules is used instead of this flat score value.
 
 ### response _ip..._
 
-**New in 0.8**
-
 Defines per-response-code rules for scoring and custom messages. This is useful
 for combined DNSBLs like Spamhaus ZEN that return different codes for different
 listing types.
--- a/docs/reference/checks/dnsbl.md
+++ b/docs/reference/checks/dnsbl.md
@@ -29,6 +29,30 @@ check.dnsbl {
         mailfrom yes
         score 1
     }
+    
+    # Example with per-response-code scoring (new in 0.8)
+    zen.spamhaus.org {
+        client_ipv4 yes
+        client_ipv6 yes
+        
+        # SBL - Spamhaus Block List (known spam sources)
+        response 127.0.0.2 127.0.0.3 {
+            score 10
+            message "Listed in Spamhaus SBL. See https://check.spamhaus.org/"
+        }
+        
+        # XBL - Exploits Block List (compromised hosts)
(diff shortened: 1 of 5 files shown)


internal/check/dnsbl/dnsbl.go
```diff
@@ -126,7 +134,9 @@ func (bl *DNSBL) readListCfg(node config.Node) error {
 	cfg.Bool("mailfrom", false, defaultBL.EHLO, &listCfg.MAILFROM)
 	cfg.Int("score", false, false, 1, &listCfg.ScoreAdj)
 	cfg.StringList("responses", false, false, []string{"127.0.0.1/24"}, &responseNets)
-	if _, err := cfg.Process(); err != nil {
+	cfg.AllowUnknown()
```
foxcpp: Preferable way to add configuration directives with complex parsing is via `cfg.Callback`. 
See internal/table/static.go for an example.

internal/check/dnsbl/common.go
```diff
@@ -113,6 +119,56 @@ func checkIP(ctx context.Context, resolver dns.Resolver, cfg List, ip net.IP) er
 		return err
 	}
 
+	// If ResponseRules is configured, use new behavior
+	if len(cfg.ResponseRules) > 0 {
```
foxcpp, line 184: This might be more readable if extracted into a separate function. On top of that, should we also provide similar functionality for checkDomain?

docs/reference/checks/dnsbl.md
```diff
@@ -171,3 +195,55 @@ will be rejected.
...
 and override results of other blocklists.
+
+**Note:** When using `response` blocks (see below), the score from matching response
+rules is used instead of this flat score value.
+
+---
+
+### response _ip..._
+
+**New in 0.8**
```
foxcpp: We don't add annotations like that. Besides, it is incorrect - the first release this change may end up in is 0.9.

internal/check/dnsbl/dnsbl.go
```diff
@@ -144,6 +154,19 @@ func (bl *DNSBL) readListCfg(node config.Node) error {
 		listCfg.Responses = append(listCfg.Responses, *ipNet)
 	}
 
+	// Parse response blocks
+	for _, child := range unknown {
```
foxcpp: There should be warning in log for the user that adding `response` disables `responses` directive if both are specified.

internal/check/dnsbl/common.go
```diff
@@ -113,6 +119,56 @@ func checkIP(ctx context.Context, resolver dns.Resolver, cfg List, ip net.IP) er
...
+					}
+				}
+			}
+		}
+
+		if totalScore == 0 {
+			return nil
+		}
+
+		// Attempt to extract explanation string from TXT records
```
foxcpp: This code duplicates the LookupTXT below, I would like to avoid that if possible.

foxcpp:
Overall LGTM, fix formatting/imports and we will merge this.

foxcpp:
Thanks!

Issue #820, opened 2026-03-12
Hardcoded actions in rspamd module.


cxvqo:
The `reject` and `soft reject` actions are hardcoded to reject the message.
https://github.com/foxcpp/maddy/blob/813964dfe9191d863edfd3c0530e0f2fb93b703d/internal/check/rspamd/rspamd.go#L322
https://github.com/foxcpp/maddy/blob/813964dfe9191d863edfd3c0530e0f2fb93b703d/internal/check/rspamd/rspamd.go#L333
It would be better if it was made configurable for what to do like `ioErrAction` and `errorRespAction` are for example.
_(I personally want to quarantine the messages)_

cxvqo:
#821 Fix.

PR #821, opened 2026-03-13
Make reject and soft reject have configurable action in rspamd module.

cxvqo:
So I originally make issue #820, but I decided to do it myself.
In the rspamd check it adds 2 new action config: `reject_action` and `soft_reject_action`.
Default is kept as is, so nothing breaks or changes for anyone else.
Please notify if I do something wrong, it looks correct to me atleast.


foxcpp:
Thanks!

PR #830, opened 2026-03-26
rspamd: fix panic on unspecified tls_client

oidq:
Upgraded maddy on my server from v0.8.2 -> v0.9.1 and it immediately crashed on startup with following error:

```
panic: reflect.Set: value of type tls.Config is not assignable to type *tls.Config
```

Tried this patch on my setup and it resolved the issue. I suppose it is connected to https://github.com/foxcpp/maddy/commit/a90af91150296e9d29c991a2bbb1de0b6e6e51b5.


foxcpp:
🤦

PR #839, opened 2026-04-16
smtp: add LOGIN SASL auth directive

osnabrugge:
## Summary
Add support for a `login` SASL authentication directive in the SMTP target.

## Changes
- Extend the SMTP SASL directive parser to accept `login` alongside `plain`
- Use `sasl.NewLoginClient(...)` when `login` is selected
- Add a focused unit test covering the LOGIN directive and mechanism name

## Why
Some SMTP relays only support the `LOGIN` authentication mechanism and reject `PLAIN`. This makes Maddy usable with those upstream relays without requiring downstream patching or custom images.

## Validation
- Patch applies cleanly on current upstream
- `go test ./internal/target/smtp -run TestSASL_Login_Directive -count=1 -timeout 90s`

## Notes
This is a narrow change limited to the SMTP target SASL directive path and preserves existing `plain` behavior.

### DIFF
--- a/internal/target/smtp/sasl.go
+++ b/internal/target/smtp/sasl.go
@@ -57,12 +57,18 @@ func saslAuthDirective(_ *config.Map, node config.Node) (interface{}, error) {
 			}
 			return sasl.NewPlainClient("", msgMeta.Conn.AuthUser, msgMeta.Conn.AuthPassword), nil
 		}, nil
-	case "plain":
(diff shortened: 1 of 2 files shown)


foxcpp:
Thanks!

osnabrugge:
You're more than welcome! This was needed so that Maddy could authenticate to Azure Communication Services

PR #845, opened 2026-05-17
target.smtp: treat downstream DNS errors as temporary

ctr49:
DNS resolution failures for explicitly configured downstream SMTP/LMTP
targets are transport/infrastructure failures and should not make queued
messages permanently undeliverable.

Keep the existing smtpconn default behavior for other users such as
target.remote, but allow target.smtp/target.lmtp to mark DNS lookup
failures as 451 4.4.4 so target.queue can retry delivery.


foxcpp:
Fair enough. Thanks!

PR #848, opened 2026-05-20
fix(systemd): report READY=1 after reload

oidq:
SystemD would report maddy in "reloading (reload-notify)" state even after successful reload. It should send "READY=1" after finishing the reload.


foxcpp:
Thanks!

PR #850, opened 2026-05-22
feat: added remote syslog logging and log annotation (domain+msgid)

terminar:
This PR extends 
1. the integrated local syslog functionality to use a remote syslog
2. extends the logging with annotation for domain + msgid (if possible in the pipeline step)

Explanation:
I use Maddy in some sort of shared environment and I needed to separate log access to specific domains.
This PR combined with a filtering/routing syslog daemon (another project) fits the need.

I am not sure if this is interesting in any case but I wanted to share and give back code so here it is.

### DIFF
--- a/framework/log/log.go
+++ b/framework/log/log.go
@@ -51,6 +51,11 @@ type Logger struct {
 	// Additional fields that will be added
 	// to the Msg output.
 	Fields map[string]interface{}
+
+	// LogFields are rendered as [key=value ...] before the message text so
+	// that individual mail flows can be identified and filtered with grep.
+	// Use With() to populate this.
+	LogFields []string
 }
 
 func (l *Logger) Zap() *zap.Logger {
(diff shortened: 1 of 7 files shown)


framework/log/log.go
```diff
@@ -51,6 +51,11 @@ type Logger struct {
 	// Additional fields that will be added
 	// to the Msg output.
 	Fields map[string]interface{}
+
+	// LogFields are rendered as [key=value ...] before the message text so
+	// that individual mail flows can be identified and filtered with grep.
+	// Use With() to populate this.
+	LogFields []string
```
foxcpp, line 58: I'm not sure if this is a good solution since we already have Fields.

This is probably better done by customizing formatting at Logger level - msg_id is already added explicitly for most log records that are related to message delivery.

This, in turn, would require some further design work as current logger implementation is very simple and lacks extended configuration parsing.

Is there a specific reason why need fields formatted in this exact way?

PR #860, opened 2026-07-16
modify/dkim: Do not oversign Sender header filed by default

### LINKED ISSUE #768: Don't DKIM-oversign Sender header field
Sender header field is oversigned even though it is not documented. To make maddy more compatible with the mailing list, it should be signed only if it is included and NOT oversigned.

linsein:
fix: #768


foxcpp:
Thanks!

